Issue #136, opened 2016-12-25
Create live blogging support to Articles


veganstraightedge:
Here's what I'm thinking.

- [x] Add something like a `parent_id` or `collection_id` to Article. Defaults to `nil`.
- [x] If an Article has children Articles, render the collection of them inside of `.e-content` in reverse chronological order (latest on top) with fields: 
  - [x] title
  - [x] subtitle
  - [x] photo 
  - [x] content
  - [x] timestamp
  - [x] contributions

PR #161, opened 2016-12-31
Article Collections / Collection posts

bookis:
This is a first draft of having the idea of Article Collections (#136). Some notes:

- I've made the naming `collection` for the parent and `collection_posts` for children.
- I've hidden all children from any `index` view
- I've added a column in the `admin/articles/index` which shows how many children a collection has
- When viewing `admin/articles/show` the same table  as :point_up: will show for the children
- Also in :point_up: there is a button to add to the collection (unless it's already a child, no inception collections)
- Draft/published rules work the same for children, visible in draft mode, but need to be marked as published for prod.
- Ordered chronologically by `published_at`.
- Added some basic specs and a script to watch files for testing.

Open to suggestions on UX and naming. Bueller?


bookis:
Ok. I think this is ready for review. 

- I implemented websockets but there is one major flaw, articles can be set to publish in the future, so that isn't really an event, so we'd have to do a lot of work to make a system that correctly syncs posts with websockets. That is in a different branch if anyone wants it.
- Instead I implemented a simple polling thing.
- Only on articles that already have at least one child post published in the last two days
- The page will poll every minute
- A floating notice banner says "PSST! REFRESH FOR N NEW POSTS"

app/helpers/articles_helper.rb
```diff
@@ -1,5 +1,16 @@
 module ArticlesHelper
 
+  def article_tag(article, &block)
+    klasses = ["h-entry"]
+    klasses << "article-with-no-header-image" if article.image.blank?
+
+    # Data attributes are used to determine how the article should be polled for updates
+    data = {id: article.id, published_at: Time.now.to_i}
```
veganstraightedge: Why is this `Time.now` instead of`article.published_at`?

app/models/article.rb
```diff
@@ -7,13 +7,20 @@ class Article < ApplicationRecord
...
 
   scope :draft,       -> { where(status: Status.find_by(name: "draft")) }
   scope :edited,      -> { where(status: Status.find_by(name: "edited")) }
   scope :designed,    -> { where(status: Status.find_by(name: "designed")) }
   scope :published,   -> { where(status: Status.find_by(name: "published")) }
 
   scope :live,        -> { where("published_at < ?", Time.now) }
+  scope :recent,      -> { where("published_at BETWEEN ? AND ?", Time.now - 2.days, Time.now) }
+
+  scope :chronological, -> { order(published_at: :desc) }
```
veganstraightedge: Master has a `default_scope` for this.

https://github.com/crimethinc/website/blob/master/app/models/article.rb#L26

app/views/admin/articles/new.html.erb
```diff
@@ -1,2 +1,11 @@
-<%= render 'admin/page_header', thing: @article %>
+<% if @collection %>
+  <h1>
+    <%= link_to thing_type(@collection).pluralize, [:admin, thing_type(@collection).pluralize.downcase] %>
```
veganstraightedge: What possible types could this collection ever be? 
Is this future planning or are there multiple types now now?

config/routes.rb
```diff
@@ -25,6 +25,8 @@
   # Articles Atom Feed
   get "feed", to: "articles#index", defaults: { format: "atom" }, as: :feed
 
+  # Articles - Collection Items
+  get "articles/:id/collection_posts", to: "collection_posts#index"
```
veganstraightedge: I don't love `collection_posts` in the URL. But I'm also confused why this route is. 

Is this different than the route for an `Article`?

app/helpers/articles_helper.rb
bookis: I'll be checking for `published_at` on or after this time. Not the original time of the published post. If the root post was published 24 hours ago, and 10 new posts since, those 10 new posts are already on the page, I don't want to poll for those 10 new posts. Only posts that have been published after `Time.now`.

app/models/article.rb
bookis: Just saw this, so I haven't corrected in the rebase.

app/views/admin/articles/new.html.erb
bookis: Just `Article`.

PR #206, opened 2017-01-12
Rename crimethinc/magazine to crimethinc/website in README

guitsaru:


veganstraightedge:
Good catch, @guitsaru. Thanks!

Issue #31, opened 2016-08-30
Add atom/rss feeds


veganstraightedge:
## RSS

- [x] /podcast gets an RSS feed of just the podcast
  - Copy this RSS feed and templatize it
  - https://crimethinc-staging.herokuapp.com/podcast/feed

## ATOM

- [x] Article subset URLs get the Articles feed template with a subset of Article.live.published
  - [x] All articles in a Category
  - [x] All articles on a Tag
  - [x] All articles on a Theme
  - [ ] All articles by an Author (not built yet)

- [x] All other URLs get an Atom feed of all `Article.live.published`

- [x] Add auto discover `feed:` URL to `<head>`

See my other repo's ATOM feed for an example. You can use `builder` still though.

https://github.com/darkmatterapp/publisher-server-rails/blob/master/app/views/posts/index.atom.erb

PR #225, opened 2017-01-14
Add Podcast Feed

guitsaru:
The only problem I ran into while doing this was that the iTunes categories aren't set up to match how they appear in the feed.  We have them saved as a comma separated list, but the feed has Philosophy nested under Society & Culture.  For now they're all top level category tags.

``` xml
<itunes:category text="News &amp; Politics"/>
<itunes:category text="Society &amp; Culture">
  <itunes:category text="Philosophy"/>
</itunes:category>
<itunes:category text="Arts"/>
```

#31

### DIFF
--- a/app/views/podcast/feed.rss.builder
+++ b/app/views/podcast/feed.rss.builder
@@ -35,9 +35,12 @@ xml.rss "version"      => "2.0",
       xml.text! @podcast.tags
     end
 
-    @podcast.itunes_categories.split(",").map(&:strip).each do |category|
-      xml.tag!("itunes:category", text: category)
+    # TODO: These are hardcoded for now.  Find a simple way to nest categories.
+    xml.tag!("itunes:category", text: "News &amp; Politics")
+    xml.tag!("itunes:category", text: "Society &amp; Culture") do
+      xml.tag!("itunes:category", text: "Philosophy")
     end
+    xml.tag!("itunes:category", text: "Arts")
 
     xml.tag!("itunes:image", href: @podcast.image)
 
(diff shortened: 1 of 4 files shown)


veganstraightedge:
That's true. Good eye. I struggled with what to do about that when building the Podcasts CRUD. But it was crunch time, so I punted.

Since we only have one podcast, you can just hardcode that bit for the iTunes categories. We'll burn that bridge later if we ever have a second podcast. Cool?

guitsaru:
Ok, I've hardcoded it for now.

app/views/podcast/feed.rss.builder
```diff
@@ -0,0 +1,92 @@
...
+    end
+    xml.tag!("itunes:author") do
+      xml.text! @podcast.itunes_author
+    end
+    xml.tag!("itunes:keywords") do
+      xml.text! @podcast.tags
+    end
+
+    # TODO: These are hardcoded for now.  Find a simple way to nest categories.
+    xml.tag!("itunes:category", text: "News &amp; Politics")
```
veganstraightedge: Looks like `builder` is html encoding the `text:` string. 

This
`"News &amp; Politics"`

is becoming this:
`"News &amp;amp; Politics"`

I think you can just use `"News & Politics"`

app/views/podcast/feed.rss.builder
```diff
@@ -0,0 +1,92 @@
...
+    xml.tag!("itunes:category", text: "Society &amp; Culture") do
```
veganstraightedge: Same thing here.

PR #228, opened 2017-01-15
Add atom feeds for articles, tags and categories

guitsaru:
screen shot 2017-01-14 at 22 55 19

#31

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -2,11 +2,20 @@ class CategoriesController < ApplicationController
 
   def show
     @category = Category.find_by!(slug: params[:slug])
-    @articles = @category.articles.page(params[:page]).per(15)
+    @articles = @category.articles.live.published.page(params[:page]).per(15)
 
     @html_id  = "page"
     @body_id  = "category"
     @title    = @category.name
+  end
+
+  def feed
+    @category = Category.find_by!(slug: params[:slug])
+    @articles = @category.articles.live.published.page(params[:page]).per(25)
+
+    @title    = @category.name
+
+    render "articles/index"
   end
 
 end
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -42,6 +42,13 @@ def icon_sizes
     [57, 60, 70, 72, 76, 114, 120, 144, 150, 152, 180, 192, 200, 300, 310, 500, 600]
   end
 
+  def largest_touch_icon_url
+    size       = icon_sizes.last
+    dimensions = "#{size}x#{size}"
+
+    asset_url("icons/icon-#{dimensions}.png")
+  end
+
```
veganstraightedge: Very clever. Good thinking on this.

app/views/shared/head/_link_tags.html.erb
```diff
@@ -1,5 +1,9 @@
 <!-- Backward compatibility: RSS/Atom -->
-<%= auto_discovery_link_tag(:atom, feed_path, title: setting(:page_title)) %>
+<% if content_for?(:auto_discovery) %>
+  <%= yield :auto_discovery %>
+<% else %>
+  <%= auto_discovery_link_tag(:atom, feed_path, title: setting(:page_title)) %>
+<% end %>
```
veganstraightedge: 👍🏻

app/controllers/categories_controller.rb
```diff
@@ -2,11 +2,20 @@ class CategoriesController < ApplicationController
 
   def show
     @category = Category.find_by!(slug: params[:slug])
-    @articles = @category.articles.page(params[:page]).per(15)
+    @articles = @category.articles.live.published.page(params[:page]).per(15)
```
veganstraightedge: Good catch adding the `.live.published` to these collections. We don't currently have any unpublished drafts. But this would've allowed them through. 🌟

PR #234, opened 2017-01-17
Add Video pages

### LINKED ISSUE #178: Add video permalinks
- [x] /watch shows all videos

- [ ] /watch/slug should show just that one video at full size

guitsaru:
image

Closes #178

### DIFF
--- a/app/controllers/videos_controller.rb
+++ b/app/controllers/videos_controller.rb
@@ -1,58 +1,9 @@
 class VideosController < ApplicationController
-  before_action :set_video, only: [:show, :edit, :update, :destroy]
-
-  # GET /videos
-  def index
-    @videos = Video.all
-  end
-
-  # GET /videos/1
   def show
+    @video = Video.find_by(slug: params[:slug])
+
+    @html_id  = "page"
+    @body_id  = "video"
+    @title    = @video.title
   end
-
-  # GET /videos/new
-  def new
-    @video = Video.new
-  end
-
-  # GET /videos/1/edit
-  def edit
-  end
-
-  # POST /videos
-  def create
-    @video = Video.new(video_params)
-
-    if @video.save
-      redirect_to @video, notice: 'Video was successfully created.'
-    else
-      render :new
-    end
-  end
-
-  # PATCH/PUT /videos/1
-  def update
-    if @video.update(video_params)
-      redirect_to @video, notice: 'Video was successfully updated.'
-    else
-      render :edit
-    end
-  end
-
-  # DELETE /videos/1
-  def destroy
-    @video.destroy
-    redirect_to videos_url, notice: 'Video was successfully destroyed.'
-  end
-
-  private
-    # Use callbacks to share common setup or constraints between actions.
-    def set_video
-      @video = Video.find(params[:id])
-    end
-
-    # Only allow a trusted parameter "white list" through.
-    def video_params
-      params.require(:video).permit(:title, :subtitle, :content, :tweet, :summary, :tweet, :summary, :image, :image_description, :content_format, :quality, :slug, :year, :month, :day, :duration, :published_at, :vimeo_id)
-    end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/videos_controller.rb
```diff
@@ -1,58 +1,9 @@
...
-  private
-    # Use callbacks to share common setup or constraints between actions.
-    def set_video
-      @video = Video.find(params[:id])
-    end
-
-    # Only allow a trusted parameter "white list" through.
-    def video_params
-      params.require(:video).permit(:title, :subtitle, :content, :tweet, :summary, :tweet, :summary, :image, :image_description, :content_format, :quality, :slug, :year, :month, :day, :duration, :published_at, :vimeo_id)
-    end
```
veganstraightedge: This was a good catch, deleting all the CRUD from the public site. It was sloppy of me to leave it behind.

Though, we should have it over in the admin section. I'll make a separate issue for that rather than growing the scope of this one.

app/views/about/watch.html.erb
```diff
@@ -12,22 +12,7 @@
...
-
-              <div class="e-content">
-                <p class="video-info">
-                  <span class="video-quality"><%= video.quality %></span>
-                  <span class="video-duration"><%= video.duration %></span>
-                </p>
-
-                <%= simple_format render_content(video) %>
-              </div>
+              <%= render "videos/video", video: video, linked: true %>
```
veganstraightedge: I built the `titles` partial with `linked: true` because it's used in several places and it was hard to infer when to link or not. But with this partial, I think you can do without that.

And `render "videos/video", video: video` can turn into `render video`. See next comment about the partial contents.

app/views/videos/_video.html.erb
```diff
@@ -0,0 +1,16 @@
+<h1 class="p-name"><%= link_to_if linked, video.title, video_path(video.slug) %></h1>
```
veganstraightedge: And here, you can probably do a `link_to_unless_current`.

app/views/videos/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<article class="h-entry" id="video-<%= @video.id %>">
+  <%= render "video", video: @video, linked: false %>
```
veganstraightedge: Same as above.

`render video` will probably get you what you want after changing the `link_to` in the partial.

config/routes.rb
```diff
@@ -80,6 +80,10 @@
   get "books/:slug/extras",      to: "books#extras",          as: :book_extras
 
 
+  # Videos
+  get "watch/:slug", to: "videos#show", as: :video
```
veganstraightedge: I think we should:

- [ ] add `/videos` to the routes
- [ ] move the contents of of `about#watch` to `videos#index`
- [ ] move the contents of of `about/index` to `videos/index`
- [ ] add a `redirect` route from `watch` to `videos` (like `listen` to `podcast`)

app/views/videos/_video.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
```
veganstraightedge: And can you add a crumb trail nav to `videos`, like the book pages have to `books`?

screen shot 2017-01-17 at 8 06 48 am

guitsaru:
All good feedback.  I took the opportunity to refactor it a bit to reuse existing elements, add pagination, etc... Let me know if there's anything else.

I'll get started on the video admin section.

guitsaru:
@veganstraightedge I just noticed that videos have year/month/day like articles.  Do we want to incorporate that into the route?

veganstraightedge:
Nah. For now, they'll just be like `/books`. In the future, if we do more recurring videos (like a video podcast / youtube channel kinda thing), we might use `/YYYY/MM/DD` routes for those videos. But for now, you can ignore those attrs.

PR #235, opened 2017-01-17
Adding short_path to articles + create redirect on save

greggawatt:


greggawatt:
@veganstraightedge I'm a little confused at what redirect is doing, If i try to go to the short_path it does the following in development.log 

`Started GET "/no-mgmt" for 127.0.0.1 at 2017-01-16 20:59:23 -0800
Processing by ArticlesController#show as HTML
  Parameters: {"short_path"=>"no-mgmt"}
  Redirect Load (0.7ms)  SELECT  "redirects".* FROM "redirects" WHERE "redirects"."source_path" = $1 ORDER BY "redirects"."id" DESC LIMIT $2  [["source_path", "/no-mgmt"], ["LIMIT", 1]]
  Redirect Load (0.5ms)  SELECT  "redirects".* FROM "redirects" WHERE "redirects"."source_path" = $1 ORDER BY "redirects"."id" DESC LIMIT $2  [["source_path", "/no-mgmt/"], ["LIMIT", 1]]
  Setting Load (0.3ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."slug" = $1 LIMIT $2  [["slug", :pinned_to_site_top_page_id], ["LIMIT", 1]]
  Setting Load (0.2ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."slug" = $1 LIMIT $2  [["slug", :pinned_to_footer_top_page_id], ["LIMIT", 1]]
  Setting Load (0.2ms)  SELECT  "settings".* FROM "settings" WHERE "settings"."slug" = $1 LIMIT $2  [["slug", :pinned_to_footer_bottom_page_id], ["LIMIT", 1]]
  Article Load (1.0ms)  SELECT  "articles".* FROM "articles" WHERE "articles"."year" IS NULL AND "articles"."month" IS NULL AND "articles"."day" IS NULL AND "articles"."slug" IS NULL ORDER BY published_at DESC LIMIT $1  [["LIMIT", 1]]
Redirected to http://localhost:3000/
Completed 302 Found in 10ms (ActiveRecord: 2.8ms)`

Any ideas?

veganstraightedge:
Redirects are checked in front of each page load.

They're checked for `source_path` matching the `request.path`. 

If a match is found, then the render chain is interrupted and and the request is redirect to that redirect record's `target_path`.

***

Did you create a `Redirect` with `source_path` and `target_path` in you dev db?

veganstraightedge:
https://github.com/crimethinc/website/blob/master/app/controllers/application_controller.rb#L79-L89

greggawatt:
I didn't, I expected it to follow the new short_path i defined in the article edit, but instead it's just rendering the home page.

greggawatt:
Hmm doesn't seem like redirects are the issue, i'll dive deeper

veganstraightedge:
Redirecting to home is the fallback last ditch option if no other page / route is found to match.

*** 

Redirection *could* look at Article's and their `short_path`, but I think for user expectation and clarity, I think it'd be better if all of the redirect pairs (source and target) are in one place.

app/models/article.rb
```diff
@@ -35,7 +35,9 @@ def name
   end
 
   def path
-    if published?
+    if published? && short_path
```
greggawatt: @veganstraightedge question here, Which URL should take precedence? I'm making a decision right now that short_path takes over if it exists and generates the paths, but i'm unsure this is correct.

veganstraightedge: `short_path` shouldn't be in play here. 

`path` is the *canonical` path/url for an Article/Page/etc.

`short_path` should only be used to create Redirects.

***

Sorry that there aren't better specs / definition around this feature (and others). I've mostly been working alone and have too much tribal knowledge in my brain and not enough documentation. Classic. 

If you stick around for a while, you'll get a better lay of the land. And hopefully we can get better specs and docs written.

veganstraightedge: I think `short_path` can be as simple as just an attr on Article.

Then Article can have an `after_save` that creates a Redirect using `short_path` and `path`.

Make sense?

greggawatt: Ah, yes so it's not replacing but just being an automator of sorts for making redirects based on the field. Got it.

greggawatt: And no worries about docs, I understand. This codebase is really well written, I feel like i'm going to learn a lot helping out 💯

greggawatt:
Hey sorry for the delay on this, Inauguration excitement got me off the computer. I think I got this, mind giving feedback @veganstraightedge?

app/controllers/admin/articles_controller.rb
```diff
@@ -89,6 +89,8 @@ def article_params
                                     :published_at, :tags, :categories,
                                     :image, :image_description, :css, :hide_layout,
                                     :header_background_color, :header_text_color,
+									                  :short_path,
+                                    contributions_attributes: [
```
bookis: `contribution_attributes` is listed twice now, and nested incorrectly. I'm guessing this is the result of a bad merge conflict.

app/models/article.rb
```diff
@@ -105,4 +108,10 @@ def generate_published_dates
   def downcase_content_format
     self.content_format = self.content_format.downcase
   end
+
+  def create_redirect
+    unless Redirect.where(source_path: short_path, target_path: path).present?
```
bookis: Here you can do `unless Redirect.exists?(source_path: short_path, target_path, path)`

app/validators/article_validator.rb
```diff
@@ -0,0 +1,10 @@
+class ArticleValidator < ActiveModel::Validator
+  def validate(record)
+    unless Article.find_by(short_path: record.short_path, id: record.id)
```
bookis: Why `find_by` `id` and `short_path`? The next find also finds by `short_path` and you already know the article with that `id`, it's `record`. Unless there is some nuance I'm missing here I'm guessing you could get away with just `validates :short_path, uniqueness: true` on `article.rb` and not have this custom validator. You could also add a similar validator on `Redirect` and handle the error `create_redirect`if it fails to save.

app/models/article.rb
```diff
@@ -21,6 +22,8 @@ class Article < ApplicationRecord
   before_validation :generate_published_dates, on: [:create, :update]
   before_validation :downcase_content_format,  on: [:create, :update]
 
+  after_save :create_redirect
```
bookis: I'm guessing the `Redirect` being created is mandatory, I would think doing a `before_save` would be better. If the `Redirect` fails to save with the `after_save`, the `Article` has already saved with a short path which doesn't have a redirect.

app/controllers/admin/articles_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Admin::ArticlesController < Admin::AdminController
   before_action :authorize
   before_action :set_article,              only: [:show, :edit, :update, :destroy]
-  before_action :set_contribution_options, only: [:new, :edit]
+  before_action :set_contribution_options, only: [:new, :edit, :update]
```
bookis: Since `update` doesn't use the contributions data unless it fails to save (and render `:edit`), may as well not do the finds. You could just manually call `set_contribution_options` down on line 47 before `render :edit`.

app/validators/article_validator.rb
greggawatt: So i'm not understanding the logic around how to make the validation on redirect know if it exists on the article. Any thoughts on that?

veganstraightedge:
Hey, let's scale this back to get it landed on master sooner than later.

Instead of all of the other things, for now, just add `short_path` to Article.
Don't validate it and don't create a Redirect. We can move those things into separate issues.

Cool?

veganstraightedge:
Fewer long running branches is better.

greggawatt:
I think I solved the issues if you want it as is.

greggawatt:
Do you still want me to split it up?

veganstraightedge:
Yes, please.

PR #240, opened 2017-01-18
Refactor Article and Page with concerns

### LINKED ISSUE #72: Refactor Page/Article models
- DRY
- import a Post module

guitsaru:
Also added some concerns for some other heavily duplicated code: slug generation and name from title and subtitle.

Closes #72

### DIFF
--- a/app/models/concerns/post.rb
+++ b/app/models/concerns/post.rb
@@ -1,0 +1,61 @@
+module Post
+  extend ActiveSupport::Concern
+
+  include NameFromTitle
+  include Slug
+
+  included do
+    belongs_to :user, optional: true
+    belongs_to :status
+
+    scope :draft,       -> { where(status: Status.find_by(name: "draft")) }
+    scope :edited,      -> { where(status: Status.find_by(name: "edited")) }
+    scope :designed,    -> { where(status: Status.find_by(name: "designed")) }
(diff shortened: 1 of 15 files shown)


app/models/concerns/post.rb
```diff
@@ -0,0 +1,61 @@
+module Post
+  extend ActiveSupport::Concern
+
+  include NameFromTitle
+  include Slug
```
veganstraightedge: My only small issue with this is the naming. It feels inconsistent.

Feels like it should be either:
- `Name` and `Slug` or
- `NameFromTitle` (though technically it's from title and subtitle) and `SlugX` (not sure what X would be though)

Is there a pattern or naming convention for things like this?

guitsaru: Yeah.  I started with just `Name`, but changed it when lots of models implemented name differently.  I'm fine with changing it back, but wanted to make sure it wasn't confusing.

guitsaru: Maybe `Nameable` and `Sluggable` are closer to the actual convention, but then `Post` doesn't quite fit.

guitsaru: I've changed it back to `Name` for now.  I think that's the simplest option, but it's easy to change later.

PR #246, opened 2017-01-21
Test Coverage

guitsaru:
Current Status:

screen shot 2017-01-24 at 21 32 11


veganstraightedge:
Thanks for taking this on. I know it's a big gnarly task and that I was a bad kid in not writing specs before now. You're a real trooper, @guitsaru.

guitsaru:
@veganstraightedge not a problem; happy to help.

`SimpleCov` now says 100% covered, but that's not completely accurate.  Next step is to go through and write tests for the controllers that haven't been hit yet.

veganstraightedge:
You're gonna remove `WIP` from the PR title when it's ready for review, right?

guitsaru:
@veganstraightedge yep

guitsaru:
@veganstraightedge I think I've got a while to go yet. It might be better to review what's there now to get it in early.  Also it's going to be a huge PR, so splitting it up would be good for review.

Let me know what you think and I can remove the `WIP` now.

config/routes.rb
```diff
@@ -136,7 +136,7 @@
 
 
   # Misc plumbing infrastructure
-  get "manifest.json",  to: "misc#mainfest_json"
+  get "manifest.json",  to: "misc#manifest_json"
```
veganstraightedge: Great catch.

veganstraightedge:
@guitsaru Since this PR is gonna end up being so big, how about you break it up into smaller chunks? What you've got so far as the first chunk. Then each subsequent PR of specs as it makes sense too.

guitsaru:
@veganstraightedge sounds good.

app/helpers/admin_helper.rb
```diff
@@ -1,5 +1,5 @@
 module AdminHelper
-  def thing_type(thing)
-    request.path.split('admin/').last.split("/").first.capitalize.singularize
+  def thing_type(_thing)
```
bookis: Any reason you didn't just get rid of `_thing`? And maybe rename to `current_resource_name` or something.

app/helpers/application_helper.rb
```diff
@@ -67,6 +67,6 @@ def br_to_p(html)
   end
 
   def homepage?
```
bookis: Looks like this method is never used. It can be removed.

guitsaru:
@bookis: I agree about the factories, but didn't want to pull too much in without talking about it first and didn't get around to it because of how busy the weekend was.  I'll pull that in in the next phase: it'll be especially useful for the controller testing.

app/helpers/admin_helper.rb
guitsaru: At the time I was focusing more on testing than changing code, but it's a good idea to take it out.

app/helpers/application_helper.rb
guitsaru: Good call

veganstraightedge:
@guitsaru If this is a good save point, I'll acceptance test then merge PR. Are you good?

guitsaru:
@veganstraightedge go for it

Issue #71, opened 2016-09-26
Refactor tags/categories/themes


veganstraightedge:
- Into one thing
- Works with Pages and Articles

PR #258, opened 2017-01-25
Support tagging pages

withoutboats:
Per #71 

Current commit is just the data model change, I'll add UI in a subsequent commit to this PR.

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -78,7 +78,9 @@ class Admin::ArticlesController < Admin::AdminController
   end
 
   def organize_article
-    @article.save_tags!(params[:tags])
+    tag_assigner = TagAssigner.parse_glob(params[:tags])
+    tag_assigner.assign_tags_to!(@article)
+
     @article.save_categories!(params[:categories])
   end
 
--- a/app/controllers/admin/pages_controller.rb
+++ b/app/controllers/admin/pages_controller.rb
@@ -58,15 +58,14 @@ class Admin::PagesController < Admin::AdminController
   end
 
   def organize_page
-    # TODO
-    # @page.save_tags!(params[:tags])
-    # @page.save_categories!(params[:categories])
+    tag_assigner = TagAssigner.parse_glob(params[:tags])
+    tag_assigner.assign_tags_to!(@page)
   end
 
   def page_params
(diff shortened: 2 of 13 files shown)


withoutboats:
This seems ready to review. It allows admins to tag pages. Those tags don't get displayed on the page, and the pages aren't displayed on the tag's page, because I'm not sure what the UI y'all want for that is.

In terms of deprecated categories and themes, it seems totally reasonable to just migrate categories to tags, since they have the same representation. I'm not sure if all the categories that exist want to still exist as tags, though; the ones that are seeded are the old categories from the earlier site, which aren't really topical.

Themes could also be migrated, though they seem to have a constraint that an article only has one theme; I don't know if this is something we want to keep.

veganstraightedge:
Hi @withoutboats! Looks pretty good so far at quick glance.

But if I'm understanding it correctly, it's not quite ready to merge as is. Unless I'm missing something.

**Is there a way to discern that one tag is a tag and another is a category?**

***

There is no difference between the two in structure, but how we use `Category` is like sections in a newspaper. 

`Tag` is topical. There can be an _Arts_ `Article` *tagged* with "foo" and a _History_  `Article` *tagged* with "foo".

An `Article` can have many `tags`. Way more than we (currently) want to show in the header image area.

An `Article` can have many `categories`, but we keep it between 1 and 3. So we can show them in the header image area.

`Categories` are used as a top level item in the site nav (footer). `Tags` are not.

***

Am I missing how this is done?
If I'm not and it doesn't currently work that way, can you add that to your PR please?

(Sorry this Issue didn't have better specs.)

PS. `Themes` have a similar requirement. Structurally, they're very similar to `Tags` (even if we removed the one theme per `Article` restriction, which I'd be fine with.)

withoutboats:
> Is there a way to discern that one tag is a tag and another is a category?

This doesn't actually change anything about either categories or themes - this PR only allows tags to be applied to page (so articles still have both categories and tags).

The questions in my previous comments were to suss out the requirements you've enumerated, thanks :-)

If you want we can land these changes incrementally or I can keep iterating on the other changes in #71 on this branch, it doesn't make a difference to me.

veganstraightedge:
Oh great! I misunderstood. 

Incremental is better than all at once. I'll ask @bookis to review this PR too.

PR #264, opened 2017-01-27
Adds sitemap gem & config

gesa:


config/sitemap.rb
```diff
@@ -0,0 +1,42 @@
...
+      "/kickstarter/2017/"
+  ]
+
+  tce_languages = ["czech", "deutsch", "espanol", "polski", "portugues", "quebecois", "slovenscina", "slovensko"]
+  tce_languages.each do |lang|
+    static_paths.push("/tce/#{lang}/", "/tce/#{lang}/get/")
+  end
+
+  static_paths.each do |path|
+    add path, changefreq: "monthly", lastmod: "2016-01-01"
```
gesa: thank you, those helped! now only about 400 paths shy of everything.

`2016-01-01` is just the dummy date bc those aren't (likely to be) updated. i suppose it would be best to update it to 2017 tho since that was in fact the last update. good catch.

config/sitemap.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  [Article, Book, Episode, Page, Podcast, Video].each do |model|
+    model.find_each do |page|
+      add page.path, lastmod: page.updated_at
+    end
+  end
+end
```
gesa: yes! i can actually create a "sitemap index" that references all three (sitemap & two feeds) thank you

gesa:
oh also we gotta do the thing to generate it too https://github.com/kjvarga/sitemap_generator/wiki/Generate-Sitemaps-on-read-only-filesystems-like-Heroku

PR #288, opened 2017-02-12
Add Dockerfile and instructions on how to use it

mutantmonkey:
I typically do my development on Linux using Docker, and I thought I'd share this Dockerfile I wrote. Since the bootstrap script doesn't work on Linux, I'm hoping this will be a convenient way for Linux users to contribute.

I'm currently overriding the Rails database config with the `DATABASE_URL` environment variable, which prevents the `db:drop` and `db:create` commands from working, which is why I'm creating the database manually. If you'd prefer, I can have the Dockerfile update config/database.yml appropriately.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ruby:2
+
+RUN apt-get update \
+        && apt-get install -y apt-transport-https software-properties-common \
+        && add-apt-repository "deb https://cli-assets.heroku.com/branches/stable/apt ./" \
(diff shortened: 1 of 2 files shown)


steveklabnik:
I am not super great with Docker, but this looks reasonable to me 👍

veganstraightedge:
Thanks @mutantmonkey. We hope you stick around and help us build this thing. There's lots to do. If you have any questions or want to discuss priorities and interests, feel free to email me or message me in Keybase or DM me on Twitter.

PR #304, opened 2017-02-17
Extend media embed syntax for link URLs

### LINKED ISSUE #303: Extend media embed syntax for link URLs
Currently this embed code:

```
[[http://example.com Caption text.]]
```

will create this HTML:

```html
<figure>
  <img src="http://example.com" />
  <figcaption>Caption text.</figcaption>
</figure>
```

***

If you want to make the image a clickable link, you can't use the embed code and instead have to write this HTML manually:

```html
<figure>
  <a href="http://example.com/link"><img src="http://example.com" /></a>
  <figcaption>Caption text.</figcaption>
</figure>
```

***

Current idea for extending the `[[ ]]` syntax:

```
[[EMBED_URL OPTIONAL_LINK_URL Optional caption text.]]
```

This would require the implementation to inspect the second thing after space separation and see if it it looks like a URL. If so, treat it as a link and the thing after the next space as the caption. If not, treat it as a caption. Seems doable.

Where it gets more complicated is if the `EMBED_URL` or `OPTIONAL_LINK_URL` are actually paths instead of URLs. A little trickier to discern them from the optional caption text.

***

I'm open to other ideas.

guitsaru:
Closes #303.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -136,7 +136,7 @@ class ApplicationController < ActionController::Base
         embed_tag_pieces = embed_tag.split(" ")
 
         url     = embed_tag_pieces.shift
-        link    = embed_tag_pieces.shift if url_or_path?(embed_tag_pieces.first)
+        link    = embed_tag_pieces.pop if url_or_path?(embed_tag_pieces.last)
         caption = embed_tag_pieces.join(" ")
 
         expanded_embed(url, caption: caption, link: link)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -136,9 +136,10 @@ class ApplicationController < ActionController::Base
         embed_tag_pieces = embed_tag.split(" ")
 
         url     = embed_tag_pieces.shift
+        link    = embed_tag_pieces.shift if url_or_path?(embed_tag_pieces.first)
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -187,8 +188,11 @@ def expanded_embed(url, caption=nil)
       end
     end
 
-    render_to_string partial: "/articles/embeds/#{slug}.html.erb", locals: { embed_id: embed_id || url, caption: caption }
+    render_to_string partial: "/articles/embeds/#{slug}.html.erb", locals: { embed_id: embed_id || url, caption: caption, link: link }
   end
   helper_method :expanded_embed
 
+  def url_or_path?(string)
+    string.match?(/^(http|\/)\S+/)
```
veganstraightedge: What cases does this cover?

- https URLs
- http URLs
- /relative/to/site/root/paths
- relative/to/document/paths

The last one seem least important and probably the hardest to account for.

guitsaru: It covers the first 3 cases.  I can modify it to cover the final case, but that seems like it would have the most false positives.  I can see wanting to have a caption with a / in it quite often.

veganstraightedge: Agreed. I don't think the 4th option is worth it.

PR #326, opened 2017-02-27
Add id to embed tags

### LINKED ISSUE #316: Add #id to the end of [[ ]] embed syntax
Ok. Here's another (and I think the final) addition to `[[ ]]`.

The need to be able to apply an `id=...` to the outer wrapper of whatever media is embedded (`<figure>`, `<div class="video-container">`, etc).

I think we tack it on the end and then look at the last two things for URL/id formatting. Thing is both are optional and id doesn't require a link URL.

Here are some examples.

```
[[ https://example.com/image.png ]]
[[ https://example.com/image.png Optional caption. ]]

[[ https://example.com/image.png optional caption /path/to/link ]]
[[ https://example.com/image.png optional caption https://example.com/path/to/link ]]

[[ https://example.com/image.png #id ]]
[[ https://example.com/image.png optional caption #id ]]

[[ https://example.com/image.png optional caption /path/to/link #id ]]
[[ https://example.com/image.png optional caption https://example.com/path/to/link #id ]]

[[ https://example.com/image.png optional caption #id /path/to/link ]]
[[ https://example.com/image.png optional caption #id https://example.com/path/to/link ]]
```

guitsaru:
Closes #316.


app/controllers/application_controller.rb
```diff
@@ -190,11 +191,24 @@ def expanded_embed(url, caption: nil, link: nil)
...
   end
+
+  def remove_id(pieces)
+    return unless pieces.present?
+
+    id_string = pieces.detect { |piece| piece =~ /id:\S+/ }
+
+    if id_string
+      id = id_string.split(":").last
+      pieces.delete(id_string)
```
veganstraightedge: I don't understand what this line is doing. If the next line returns `id`, what does deleting any amount of `pieces` do?

guitsaru: It removes it from embed_tag_pieces so it can properly parse the link and caption.

veganstraightedge: How does it do that? `pieces` is a local var inside of this method. 
Sorry for my confusion. I just don't see it doing what you're saying it does.

guitsaru: pieces isn't a local var, it's passed by reference as an argument.  You can see a simpler version here:

```ruby
>> def remove_first(array)
>>   array.shift
>> end

>> test_array = [1, 2, 3]
=> [1, 2, 3]

>> remove_first(test_array)
=> 1

>> test_array
=> [2, 3]
```

guitsaru: But I take your point that this could be too confusing.  Would it be less so if I just did this all inline?

veganstraightedge: Nah. It's good. I am just mentally foggy right now.
Is this a case of `!` method name though?

guitsaru: I was basing it off of `Array#delete`, which isn't a `!` method, but the argument could be made either way.

veganstraightedge: I'm prolly overthinking it. Thanks for talking through it. Let's ship it. :shipit:

guitsaru: 👍

Issue #232, opened 2017-01-16
Add good title tags to all pages


veganstraightedge:

astronaut-wannabe:
what is wrong with the title tag?

for an example, [this article][0] has this title tag:

```html
<title>CrimethInc. : Interview: The Standing Rock Evictions : (Audio and Transcript)</title>
```

unless you mean there are older articles that do not have this format of title tag?

[0]:https://crimethinc.com/2017/02/28/interview-the-standing-rock-evictions-audio-and-transcript

veganstraightedge:
1. The `<title>` tag on most public pages is probably good. Though they should all be audited to make sure that they all have a line like:

https://github.com/crimethinc/website/blob/master/app/controllers/articles_controller.rb#L45

For example, `articles#index` doesn't.
https://github.com/crimethinc/website/blob/master/app/controllers/articles_controller.rb#L7-L9

2. Most of the admin pages don't have any meaningful `<title>` tag. So, my history is flooded with pages just called `CrimethInc.`. It'd be better if they had titles like `CrimethInc. Admin : Editing Article 3 "Article title : subtitle"`. Or something.

Make sense?

astronaut-wannabe:
@veganstraightedge, makes sense, thanks 👍

veganstraightedge:
@astronaut-wannabe Do you want to take on this issue?

astronaut-wannabe:
@veganstraightedge sure!

PR #338, opened 2017-03-13
[#232] add descriptive title tags to admin pages

astronaut-wannabe:
this addresses #232 

This puts in the wiring to make further changes/edits straightforward

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">&#xa0;</th>
<th scope="col" class="org-left">example</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">index</td>
<td class="org-left"><title>CrimethInc. Admin : Articles</title></td>
</tr>

<tr>
<td class="org-left">new</td>
<td class="org-left"><title>CrimethInc. Admin : New Articles</title></td>
</tr>

<tr>
<td class="org-left">show</td>
<td class="org-left">
<title>CrimethInc. Admin : Articles : "Under No Management"</title>
</td>
</tr>

<tr>
<td class="org-left">edit</td>
<td class="org-left"><title>CrimethInc. Admin : Editing article 719 "Under No Management : The NEW crimethinc.com"</title></td>
</tr>
</tbody>
</table>

### controllers edited 
- [x] admin_controller
- [x] articles_controller
- [x] books_controller
- [x] categories_controller
- [x] contributors_controller
- [x] episodes_controller
- [x] links_controller
- [x] pages_controller
- [x] podcasts_controller
- [x] redirects_controller
- [x] roles_controller
- [x] settings_controller
- [x] subscribers_controller
- [x] themes_controller
- [x] users_controller
- [x] videos_controller

### DIFF
--- a/app/controllers/admin/admin_controller.rb
+++ b/app/controllers/admin/admin_controller.rb
@@ -1,3 +1,18 @@
 class Admin::AdminController < ApplicationController
   layout "admin"
+
+  def admin_title(model=nil, keys = [])
+    return t(".#{action_name}_title", default: '') unless model.present?
+
+    translation_vars = {}
+
+    keys.each_with_object(translation_vars) do |key, hash|
+      hash[key] = model.send(key)
+    end
+
+    t(".#{action_name}_title", translation_vars)
+  rescue NoMethodError
+    logger.error "#{controller_path}:#{action_name} has an issue with the page title"
+    return ''
+  end
 end
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -7,6 +7,7 @@ class Admin::ArticlesController < Admin::AdminController
   # /admin/articles
   def index
     @articles = Article.root.includes(:collection_posts).page(params[:page])
+    @title = admin_title
   end
 
   # /admin/articles/1
@@ -15,17 +16,20 @@ class Admin::ArticlesController < Admin::AdminController
     if @article.collection_id.present?
       @collection = Article.find(@article.collection_id)
     end
+    @title = admin_title(@article, [:title, :subtitle])
   end
 
   # /admin/articles/new
   def new
     @collection = Article.find(params[:id]) if params[:id]
     @article = Article.new
+    @title = admin_title
   end
 
   # /admin/articles/1/edit
   def edit
     @collection = Article.find(@article.collection_id) if @article.in_collection?
+    @title = admin_title(@article, [:id, :title, :subtitle])
   end
 
   # /admin/articles
--- a/app/controllers/admin/books_controller.rb
+++ b/app/controllers/admin/books_controller.rb
@@ -5,19 +5,23 @@ class Admin::BooksController < Admin::AdminController
   # /admin/books
   def index
     @books = Book.page(params[:page])
+    @title = admin_title
   end
 
   # /admin/books/1
   def show
+    @title = admin_title(@book, [:title, :subtitle])
   end
 
   # /admin/books/new
   def new
     @book = Book.new
+    @title = admin_title
   end
 
   # /admin/books/1/edit
   def edit
+    @title = admin_title(@book, [:id, :title, :subtitle])
   end
 
   # /admin/books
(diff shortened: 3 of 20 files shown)


app/controllers/admin/articles_controller.rb
```diff
@@ -26,6 +28,7 @@ def new
   # /admin/articles/1/edit
   def edit
     @collection = Article.find(@article.collection_id) if @article.in_collection?
+    @title = t('.edit_title', id: @article.id, title: @article.title, subtitle: @article.subtitle)
```
astronaut-wannabe: I wanted to put a method in `admin_controller` that looked something like this:

```ruby
def admin_title(action:, model:)
  t(".#{action}_title", id: model.id, title: model.title, subtitle: model.subtitle)
end
```

and then, for example, in `admin/articles_controller` just do 
```ruby
@title = admin_title(current_action, @article)
```

but not every model uses `title` (some use `name`) so it seemed like it was more confusing than just putting this `t(...)` line wherever it is needed

astronaut-wannabe:
@veganstraightedge this is now functionally done. let me know if you have any feedback

app/controllers/admin/articles_controller.rb
astronaut-wannabe: actually, I have an idea to make it work. I will add a commit on top of this and you can see which you prefer

app/controllers/admin/admin_controller.rb
```diff
@@ -1,3 +1,19 @@
 class Admin::AdminController < ApplicationController
   layout "admin"
+
+  def admin_title(model=nil, keys = [])
+    return t('.index_title') unless model.present?
```
veganstraightedge: Which `index_title` does this refer to?

app/controllers/admin/admin_controller.rb
```diff
@@ -1,3 +1,19 @@
...
+
+    translation_vars = {}
+
+    keys.each_with_object(translation_vars) do |key, hash|
+      hash[key] = model.send(key)
+    end
+
+    t(".#{action_name}_title", translation_vars)
+  rescue NoMethodError
```
veganstraightedge: Does this rescue need a `begin` earlier in the method?

astronaut-wannabe: nope, method blocks in ruby already have an implicit exception block.

app/controllers/admin/admin_controller.rb
astronaut-wannabe: this refers to any of the `index_title`s I added in the [`en.yml`][0] file.

this is using the "[lazy lookup][1]" dot notation from the Rails i18n feature

[0]:https://github.com/crimethinc/website/pull/338/files#diff-37c9b835927e3c363588f172de9bb6faR149
[1]:http://guides.rubyonrails.org/i18n.html#lazy-lookup

veganstraightedge: I thought I understood "lazy lookup", but this confuses me.

What provides the context to instruct which `index_title` will be used when?

astronaut-wannabe: sorry, @veganstraightedge, I was using "lazy" in the "too lazy to type the whole path" sense, not "lazy" in the computer science/programming sense.

So I added a bunch of translations that look like this
```yaml
  admin:
    title_prepend: "Admin"
    articles:
      index_title: "Articles"
      show_title: "Articles : \"%{title}\""
      edit_title: "Editing article %{id} \"%{title} : %{subtitle}\""
```

one way of getting at the `index_title` in the `yaml` would be:

```ruby
t('admin.articles.index_title')
```

those paths can get pretty nested though, so rails gives you a special notation where if your translations are organized the same as your controller (or views) you can do the same lookup like this  (note the dot):

```ruby
t('.index_title')
```

this works here because the method is in the admin controller

veganstraightedge:
This looks good. One last thing.

Could you add a case for new views?

`<title>CrimethInc. Admin : New Articles</title>`

astronaut-wannabe:
@veganstraightedge update with the `/new` route titles

PR #342, opened 2017-03-16
tweak headline shadows so they aren't cut off

gesa:
resolves #325 

(…also rearranges some media queries for legibility. i tested element by element to confirm my rearranging.)

<details>
<summary>some  screenshots:</summary>

## 320px
mbmyr

## 480px
9648x

## 800px
j4xba

## 1440px
tkz9e

</details>

PR #343, opened 2017-03-16
Add shadowed class for article headers

### LINKED ISSUE #328: Move title/subtitle text-shadow to a separate class
In preparation of adding a checkbox to only add a text shadow to article title/subtitle when selected, I'd like to move the styles to their own rule with their own class. That way we can optionally add that class to the wrapping `<article>` tag to add the text-shadow when needed. And leave it off by default when we don't need it.

gesa:
resolves #328
~depends on #342 being merged~

To make use of this class, add logic to toggle a class of `shadowed` at L1 of app/views/articles/_titles.html.erb

### DIFF
--- a/app/assets/stylesheets/views/_article.scss
+++ b/app/assets/stylesheets/views/_article.scss
@@ -51,6 +51,15 @@
           @media screen and (min-width: $small-tablet) {
             padding: 50pt;
           }
+
+          &.shadowed {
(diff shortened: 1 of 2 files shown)


gesa:
due to 
> depends on #342 being merged
seriously just ignore the diff on this PR until that's 👍

Issue #341, opened 2017-03-15
Add max character counters to tweet and summary on Article


veganstraightedge:
Count down from 115 and 200.

Add Rails validation too.

astronaut-wannabe:
@veganstraightedge are you working on this already?

veganstraightedge:
I am/was not. I'm happy that you jumped on it. 👍🏻

PR #344, opened 2017-03-16
[#341] Add max character counters to tweet and summary on Article

astronaut-wannabe:
This sort of works, but three details probably need more work:

1. ~~I never know where to put javascript in a rails app~~
2. ~~should the model validation only happen on publish? (so draft articles can have tweets/summaries that are longer)~~
3. ~~there is a difference between the newlines in the textarea and in Ruby (described below) and I don't know how to handle it, so it makes the counting code a little confusing~~

### newline issue
<details>
<summary>A newline in the `textarea` will count as 2 characters on the ruby
end.</summary>

e.g.

```textarea
foo

bar
```

translates to the ruby string
```ruby
"foo
\r\n
bar"
```
~~I didn't know the right way to handle this, so I just counted every
newline in the textarea as 2 when calculating how many characters are
left.~~

~~that said, entering a newline on twitter only counts as 1 character, so maybe I should `.gsub("\r\n", "\n")` in a `before_save` action instead?~~

I added a `before_validation` hook to replace `\r\n\` with `\n`
</details>

### tweet char counter

tweet

### DIFF
--- a/app/views/admin/articles/_form.html.erb
+++ b/app/views/admin/articles/_form.html.erb
@@ -126,7 +126,8 @@
       <div class="col-xs-12 col-sm-6">
         <div class="form-group">
           <%= f.label :tweet %>
-          <%= f.text_area :tweet, class: "form-control" %>
+          <%= f.text_area :tweet, class: "form-control", data: { 'max-length': '115', 'feedback-box': 'tweet-length' } %>
+          <div id="tweet-length"></div>
           <p class="help-block">
             Maximum 115 characters to leave room for the <code>t.co</code> short URL that Twitter creates.
             Used for syndicated tweet for this Article.
@@ -137,7 +138,8 @@
       <div class="col-xs-12 col-sm-6">
         <div class="form-group">
           <%= f.label :summary %>
-          <%= f.text_area :summary, class: "form-control" %>
+          <%= f.text_area :summary, class: "form-control", data: { 'max-length': '200', 'feedback-box': 'summary-length' } %>
+          <div id="summary-length"></div>
           <p class="help-block">
             Maximum 200 characters.
             Used for page description and previews in story cards on other sites.
--- a/spec/models/article_spec.rb
+++ b/spec/models/article_spec.rb
@@ -1,6 +1,40 @@
 require "rails_helper"
 
 describe Article do
+  describe 'validations' do
+    it 'validates tweet is less than 115 characters' do
+      invalid_article = build(:article, tweet: 'a' * 116)
+      valid_article   = build(:article, tweet: 'a' * 115)
+
+      expect(valid_article).to be_valid
+      expect(invalid_article).not_to be_valid
+    end
+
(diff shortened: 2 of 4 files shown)


spec/models/article_spec.rb
```diff
@@ -1,6 +1,16 @@
...
+  describe 'validations' do
+    it 'validates tweet is less than 115 characters' do
+      invalid_article = build(:article, tweet: 'a' * 116)
+      valid_article   = build(:article, tweet: 'a' * 114)
+
+      expect(valid_article).to be_valid
+      expect(invalid_article).not_to be_valid
+    end
+  end
+
```
veganstraightedge: 👍🏻

Please add one for `summary` while you're at it.

app/views/admin/articles/_form.html.erb
```diff
@@ -126,7 +126,8 @@
       <div class="col-xs-12 col-sm-6">
         <div class="form-group">
           <%= f.label :tweet %>
-          <%= f.text_area :tweet, class: "form-control" %>
+          <%= f.text_area :tweet, class: "form-control", data: { 'max-length': '115', 'feedback-box': 'tweet_length_feedback' } %>
+          <div id="tweet_length_feedback"></div>
```
veganstraightedge: We use hyphens instead of underscores as word separators in attr values.
Please change these to that.

app/views/admin/articles/_form.html.erb
```diff
@@ -137,7 +138,8 @@
       <div class="col-xs-12 col-sm-6">
         <div class="form-group">
           <%= f.label :summary %>
-          <%= f.text_area :summary, class: "form-control" %>
+          <%= f.text_area :summary, class: "form-control", data: { 'max-length': '200', 'feedback-box': 'summary_length_feedback' } %>
+          <div id="summary_length_feedback"></div>
```
veganstraightedge: And here too.
And in your JS at the top.

veganstraightedge:
> so maybe I should .gsub("\r\n", "\n") in a before_save action instead?

Seems good. Do that. 👍🏻

veganstraightedge:
> 2. should the model validation only happen on publish? (so draft articles can have tweets/summaries that are longer)

Nah. Validated all the time is good.

veganstraightedge:
> 1. I never know where to put javascript in a rails app

Putting it in `admin.js` was a good choice. If down the road there is lots of adminy JS, we can break it into multiple `admin/*.js` files. But we'll burn that bridge when we get to there.

spec/models/article_spec.rb
astronaut-wannabe: done

app/views/admin/articles/_form.html.erb
astronaut-wannabe: done

app/views/admin/articles/_form.html.erb
astronaut-wannabe: done

Issue #163, opened 2016-12-31
Strip EXIF data from images


veganstraightedge:
Manually at first. Then automatedly, when there's an uploader.

http://superuser.com/questions/335489/how-to-strip-exif-info-from-files-in-osx-with-batch-or-command-line/335504

astronaut-wannabe:
So, I see #173 was closed but I don't know if this was ever implemented.

I recently started doing some responsive image work for another project and thought I would share how it was coming along.

I am adding it to this issue since the flow ***should*** strip exif data as part of the ImageMagick processing (but I am not 100% sure)

---

Basically, we needed to have an image for every ActiveRecord model and wanted to be able to do something like this in the view:

```html
  <img srcset="<%= model.image.large %>  1024w,
               <%= model.image.medium %> 640w,
               <%= model.image.small %>  320w"
               sizes="(min-width: 36em) 33.3vw,100vw"
               src=<%= model.image %>
               alt=<%= model.title %> />
```

I followed these three guides for how to do everything:
- [Efficient Image Resizing With ImageMagick ](https://www.smashingmagazine.com/2015/06/efficient-image-resizing-with-imagemagick/)
- [Responsive Images Done Right: A Guide To <picture> And srcset](https://www.smashingmagazine.com/2014/05/responsive-images-done-right-guide-picture-srcset/)
- [Responsive Images: If you’re just changing resolutions, use srcset.](https://css-tricks.com/responsive-images-youre-just-changing-resolutions-use-srcset/)

To automate everything within a Rails CMS flow we used the [carrierwave](https://github.com/carrierwaveuploader/carrierwave) gem to upload the images. This gem has some nice features that made everything very easy:

1. ActiveRecord Hooks
2. You can create "versions" on_save, so we can create the small/medium/large variants
3. An ImageMagick interface via [MinMagick](https://github.com/minimagick/minimagick)
4. Ability to point your file storage to remote locations (aws, google cloud, etc.)

basically, all the code that was needed to tie all of these things together was (I am not sure if all of those `yield`s are needed, I just followed the pattern from the README):

```ruby
  class ImageUploader < CarrierWave::Uploader::Base
    include CarrierWave::MiniMagick

    storage :file
    # storage :fog

    def store_dir
      "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
    end

    version :small do
      process :create_sized_image => 320
    end

    version :medium do
      process :create_sized_image => 640
    end

    version :large do
[...]

PR #347, opened 2017-03-26
Add article image upload to the article admin page

astronaut-wannabe:
This adds the ability to upload the header image for an article from
the admin page.

This will currently
- create 3 versions of the images (small,medium,large)
- use the original `article.image` as the fallback
- is configured to use AWS ~~saves file to localdisk *not cloudfront*~~

TODO:
------

- [x] currently saves uploads to local disk. integrate with cloudfront
- [ ] specs

---

issue: https://github.com/crimethinc/website/issues/163


astronaut-wannabe:
@veganstraightedge this is currently works with local  file storage. There are instructions for cloud [here][0]

this is still a work in progress, but it works enough that you could play around with it locally to see if you have any thoughts on the admin flow

[0]:https://github.com/carrierwaveuploader/carrierwave#fog

astronaut-wannabe:
sorry, @veganstraightedge, life caught up with me.

I am picking this up and looking at it again

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,23 @@
+# TODO: what happens if the ENVs are missing?
+# TODO: is ENV safe way to handle these credentials?
+CarrierWave.configure do |config|
+  # required settings
+  config.fog_provider = 'fog/aws'
+  config.fog_directory  = 'article-header-images'
```
astronaut-wannabe: this is supposed to be whatever the base S3 bucket is. maybe that is `assets`?

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  # optional settings
+  config.fog_public     = true # defaults to true
+  config.fog_attributes = { 'Cache-Control' => "max-age=#{365.day.to_i}" } # defaults to {}
```
astronaut-wannabe: maybe use [Cache-Control: immutable][0]?

[0]:https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/

db/migrate/20170326022716_add_article_image_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddArticleImageToArticles < ActiveRecord::Migration[5.0]
+  def change
+    add_column :articles, :article_image, :string
```
astronaut-wannabe: so as not to clobber the current column.

it might be good to do some sort of migration and then rename this back to just `image`

veganstraightedge: That's a good idea. 👍🏻

veganstraightedge:
@astronaut-wannabe Hey. Sorry for the super delay on this. Chemo has been distracting me. Can you fix up the merge conflicts to get it ready for merging?

I'll add some other notes to the code itself.

app/uploaders/article_image_uploader.rb
```diff
@@ -0,0 +1,53 @@
+class ArticleImageUploader < CarrierWave::Uploader::Base
+  include CarrierWave::MiniMagick
+
+  def store_dir
+    # This will be the directory path in the CMS
+    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
```
veganstraightedge: Can you change this to match our current directory structure on S3, please?

`assets/articles/YYYY/MM/DD`

app/uploaders/article_image_uploader.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def extension_whitelist
+    %w[jpg jpeg gif png]
```
veganstraightedge: Might as well add `svg` to this list too.

app/uploaders/article_image_uploader.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def filename
+    # don't try changing the name if original_filename is missing
+    return unless original_filename
+
+    # This will be the filename for all the processed/uploaded images
+    model.title.downcase.tr(' ', '_') + File.extname(original_filename)
```
veganstraightedge: We usually name our headers `header.jpg`. 

Though if there are two articles on the same day, we name them `header.jpg` and `header-2.jpg`.

Could this Do The Right Thing™ and check to see if there are other articles on that same day and adjust the name incrementer accordingly? Or is that a hard thing?

app/uploaders/article_image_uploader.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  version :large do
+    process create_sized_image: 1024
```
veganstraightedge: Our header images are 2000px wide. So, can you add a size break point for that too?

app/uploaders/article_image_uploader.rb
```diff
@@ -0,0 +1,53 @@
...
+      img.thumbnail(width)
+      img.unsharp('0.25x0.08+8.3+0.045')
+      img.dither('None')
+      img.posterize('136')
+      img.quality('82')
+      img.define('jpeg:fancy-upsampling=off')
+      img.define('png:compression-filter=5')
+      img.define('png:compression-level=9')
+      img.define('png:compression-strategy=1')
+      img.define('png:exclude-chunk=all')
```
veganstraightedge: We've found that JPGs work best for our headers and have the smallest file size. Unless this is doing some PNG magic that I don't know about and is smaller than the JPG of the same pixel size, can you default to saving as a JPG please?

app/views/admin/articles/_form.html.erb
```diff
@@ -157,8 +157,19 @@
...
         <div class="form-group">
-          <%= f.label :image, "Header Image" %>
-          <%= f.text_field :image, class: "form-control input-lg" %>
+          <p class="help-block">The article's header image is currently set to
+            <code>
+              <%= @article.article_image.url || @article.image %>
+            </code>
+          </p>
+          <%= f.label :article_image, "Header Image Upload" %>
+          <%= f.file_field :article_image, class: "form-control" %>
```
veganstraightedge: Since the `store_dir` that I want needs to know what date the article will be published on, can you hide this part of the form on /new and only show it on /edit?

app/views/articles/_image.html.erb
```diff
@@ -1,7 +1,12 @@
...
-<% end %>
+<div class="article-image">
+  <img class="u-photo"
+       alt="<%= article.image_description %>"
+       src="<%= article.article_image.url || article.image %>"
+       <% if article.article_image.url.present? %>
+         sizes="(min-width: 36em) 33.3vw,100vw"
+         srcset="<%= article.article_image.large %>  1024w,
+                 <%= article.article_image.medium %> 640w,
+                 <%= article.article_image.small %>  320w"
```
veganstraightedge: See above about headers being 2000px at their big size.

Also, I'm not up on the `sizes` and `srcset` attrs. Can you explain to me in English what is happening here?

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,36 @@
+# TODO: what happens if the ENVs are missing?
```
veganstraightedge: Blow up early.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,36 @@
...
+# TODO: is ENV safe way to handle these credentials?
```
veganstraightedge: ENV vars is a good way to deal with these credentials.

We should prolly add a sample `.env` file called `dot_env` and setup instructions in the README.
And the `dot_env` gem.

db/migrate/20170326022716_add_article_image_to_articles.rb
veganstraightedge: Smart.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,36 @@
...
+
+# TODO: staging too?
+if Rails.env.prod?
+  CarrierWave.configure do |config|
+    config.storage = :fog
+    config.enable_processing = true
+
+    # required settings
+    config.fog_provider = 'fog/aws'
+    config.fog_directory  = 'article-header-images'
```
veganstraightedge: See my comment about `store_dir` above.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,36 @@
...
+    config.fog_credentials = {
+      # required credentials
+      provider:              'AWS',
+      aws_access_key_id:     ENV['AWS_CARRIERWAVE_ACCESS_KEY_ID'],
+      aws_secret_access_key: ENV['AWS_CARRIERWAVE_ACCESS_KEY'],
+
+      # optional values
+      region:     'us-west-2', # defaults to 'us-east-1'
+      # host:     's3.example.com',             # defaults to nil
+      # endpoint: 'https://s3.example.com:8080' # defaults to nil
```
veganstraightedge: Will the uploader give us back URLs to our Cloudfront cache or to the S3 bucket?

What we get now (manually) and what we want to continue to get is:

`https://cloudfront.crimethinc.com/assets/articles/YYYY/MM/DD/header.jpg`

db/migrate/20170326022716_add_article_image_to_articles.rb
veganstraightedge: Also, I like to treat URLs as `:text` instead of `:string` in case they grow really long.

app/uploaders/article_image_uploader.rb
astronaut-wannabe: I definitely can, I just need to understand the publication flow a little more.

My current understanding is:

the `YYYY/MM/DD` path isn't set until the article is published, but the image needs to be uploaded while the article is still in `draft`

If the article is published already, this gives the right path:
```ruby
article.path.gsub(article.slug,'')
```

but if it isn't published yet, I am not sure how to *know* the path to save the image in

app/uploaders/article_image_uploader.rb
astronaut-wannabe: fixed

app/uploaders/article_image_uploader.rb
astronaut-wannabe: This is an imagemagick script I took from this article: [Efficient Image Resizing With ImageMagick][0]

This script is basically handling jpeg by default (the `quality` and `jpeg:fancy-upsampling` settings) and the png settings will only come into play if you upload a png

[0]:https://www.smashingmagazine.com/2015/06/efficient-image-resizing-with-imagemagick/

app/views/admin/articles/_form.html.erb
astronaut-wannabe: sounds good! I think this also solves the question I had above on your `store_dir` comment

app/views/articles/_image.html.erb
astronaut-wannabe: I'll do my best, but responsive images are still pretty new to me.

 <dl>
<dt>src</dt>
<dd>- the fallback if the browser doesn't recognize the <code>srcset</code> attribute</dd>
  <dt>srcset</dt>
  <dd>- takes a comma separated list of the sources we have (in this case, the versions generated by the uploader ruby code). the <code>w</code> is describing the width of the image source, in pixels.</dd>
  <dt>sizes</dt>
  <dd>- tells the browser how many pixels each source <em>needs</em> by describing the final rendered width of the image using CSS length values. Using <code>vw</code> (as opposed to <code>px</code>) allows the image to stretch. 

<br><br>

<b>I actually have a bug here.</b> I should just have `sizes="100vw"` since the header image is always 100%. 

<br>

If the image size changes at different breakpoints in the CSS, you could encode that as <code>sizes="[media query] [length], [media query] [length], etc"</code></dd>
</dl> 

The resource I used to figure this out was this article: [Responsive Images Done Right: A Guide To \<picture\> And srcset][0]

*note: using `<picture>` would be cool for the potential art-direction it enables, but seems like more work and complication to the article workflow than it's worth*

[0]:https://www.smashingmagazine.com/2014/05/responsive-images-done-right-guide-picture-srcset/

db/migrate/20170326022716_add_article_image_to_articles.rb
astronaut-wannabe: fixed

config/initializers/carrierwave.rb
astronaut-wannabe: So, I noticed you have upgraded to rails 5.1. Is ENV still the perferred way, or should I use the `rails secrets` command now? I don't know enough to know what is better/more secure

```sh
website$ rails secrets:setup --help
Usage:
  bin/rails secrets [options]

=== Storing Encrypted Secrets in Source Control

The Rails `secrets` commands helps encrypting secrets to slim a production
environment's `ENV` hash. It's also useful for atomic deploys: no need to
coordinate key changes to get everything working as the keys are shipped
with the code.

[...]
```

app/uploaders/article_image_uploader.rb
astronaut-wannabe: I am sure there is a way.

commit de53f565e6874286689454f69598b606b5d2760e added a way for the header to be named 'header-X' where X is (1..∞) based on how many articles are found published already on a given day.
    
**BUT** my concern is that this current strategy does not make sure the header is matched to the right article.
    
for example:
    
1. create 2 articles, article-1 and article-2, to be published on the same day, in that order
2. add a header to article-2 before you add one to article-1
3. now article-1 will map to header-2 and article-2 will map to header-1
    
That is a bit confusing (even typing it was confusing).

Let me think a bit more on how to make sure the order stays correct (without adding *multiple* SQL hits)

astronaut-wannabe:
> Hey. Sorry for the super delay on this. Chemo has been distracting me.

@veganstraightedge no sorrys needed! I'd say you have your priorities in the correct order

veganstraightedge:
I'll merge this this weekend after the G20 live blog. I don't want to change an UI on anyone right before a big day. Thanks for the updates!

astronaut-wannabe:
> I'll merge this this weekend after the G20 live blog

@veganstraightedge, sounds good. Let me know your thoughts on https://github.com/crimethinc/website/pull/378 as I'd like to get some feature specs written before merging in this PR

veganstraightedge:
Ping @astronaut-wannabe.

astronaut-wannabe:
@veganstraightedge ack

astronaut-wannabe:
I think the last major case to handle is what to do if you change the publication date of an article after uploading the image.

currently if you upload an image with a publication date (say, January 1, 2018) the images will be stored in S3 under `assets/articles/2018/01/01`.

If then a different article with higher priority come in and you bumped the original article's publication date (say, to January 2, 2018) the original article would still point at `assets/articles/2018/01/01` and load the new article's header image

Seems like this would be a very common scenario, so I'd like to handle it better

config/initializers/carrierwave.rb
astronaut-wannabe: done

config/initializers/carrierwave.rb
astronaut-wannabe: maybe...

I added `config.asset_host = "https://cloudfront.crimethinc.com" `, but that apparently depends on certain AWS settings according to this [Stackoverflow thread][0] 

might be easiest to just try in staging, if the `config` doesn't work I can figure out a way to get it to work

[0]:https://stackoverflow.com/questions/9956712/use-cdn-with-carrierwave-fog-in-s3-cloudfront-with-rails-3-1

astronaut-wannabe:
closing until I have more time to give this more focus

veganstraightedge:
That's probably a good idea.

I should spec it out better too before more work is done on it.

PR #368, opened 2017-05-13
Fix duplicate articles in archive.

### LINKED ISSUE #359: Why is this article appearing twice in 2000's pagination?
https://crimethinc.com/2000/09/11/veganism

On both:

https://crimethinc.com/2000/page/1
and
https://crimethinc.com/2000/page/2

guitsaru:
During the initial article import, many of the articles had the same
created_at.  It looks like the sort wasn't deterministic when only
sorting by created_at causing some articles to be duplicated when
paginating.  This commit adds a secondary sort by article title so that
it will always sort them in the same way.

Closes #359

### DIFF
--- a/app/models/archive.rb
+++ b/app/models/archive.rb
@@ -10,12 +10,13 @@ class Archive
     @year  = year
     @month = month
     @page  = page
+    @day   = day
   end
 
   def articles
     return @articles if defined?(@articles)
 
-    @articles = Article.published.live.root.all
+    @articles = Article.published.live.root.order(title: :desc)
 
     @articles = @articles.where(year:  year)  if year.present?
     @articles = @articles.where(month: month) if month.present?


app/models/archive.rb
```diff
@@ -10,12 +10,13 @@ def initialize(year:, month:, day: nil, page: 1)
...
     @month = month
     @page  = page
+    @day   = day
   end
 
   def articles
     return @articles if defined?(@articles)
 
-    @articles = Article.published.live.root.all
+    @articles = Article.published.live.root.order(title: :desc)
```
veganstraightedge: Will this order all Articles alphabetically by title even on collections that are more than just a day?

guitsaru: This will just affect the archive page.  The calendar method is what breaks it up into individual days.  The `ORDER BY` is `ORDER BY published_at DESC, "articles"."title"` so it'll still keep them in the correct published order.

image

PR #378, opened 2017-07-04
Add config for capybara and feature specs

astronaut-wannabe:
This change adds capybara and a simple feature spec as a demo

This change will add the [capybara gem][0] and set it up so that:
- by default it runs with the `:rack_test` driver (fast)
- if you add `:js` to a spec (e.g. `scenario "foo bar", :js do ...`)
  it will run with `selenium-webdriver` (slow)
- the [`database_cleaner`][1] gem will use transactions for normal specs,
  but truncation for feature specs
- the `launchy` gem is so that when you are debugging a feature spec,
  you can write `save_and_open_screenshot` and a screenshot of the page
  at that point in time will open automatically (note: you must be using
  the `:js` flag)

---

I wanted to write a feature spec for [#347][2], but noticed capybara
wasn't set up yet.

I am not sure it is worth pulling in `database_cleaner` since once
`rspec-rails` catches up with the new minitest API ([wip PR][3]), then
`database_cleaner` will no longer be needed. Maybe it is better to
just wait for that PR to merge into `rspec-rails` (it has 2 ship-its).

[0]:https://github.com/teamcapybara/capybara#drivers
[1]:https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example
[2]:https://github.com/crimethinc/website/pull/347
[3]:https://github.com/rspec/rspec-rails/pull/1813


astronaut-wannabe:
if there is already some non-capybara way of writing feature specs I missed, feel free to close this PR

veganstraightedge:
I'm not sure up on Capybara/integration tests these days.

Does what you're doing here fit into the new Rails 5.1 idea of System Tests?

http://weblog.rubyonrails.org/2017/2/23/Rails-5-1-beta1/#system-tests

veganstraightedge:
Also, @guitsaru should probably chime in here since he's written the bulk of the specs so far.

astronaut-wannabe:
@veganstraightedge @guitsaru  so my understanding of system tests in rails 5.1 is:

they use capybara in the background and it's all configured nicely out of the box (no database_cleaner gem and custom config needed!)

the issue is that this project uses rspec, and rspec-rails doesn't have a system-test api yet.

The rspec system test API is almost ready to be merged into the rspec-rails gem (that work is in the wip PR I mentioned above).

Once that PR is merged and we update the rspec-rails gem, this PR would basically be mostly the same (*I think*) except I could remove the database_cleaner stuff.

maschwenk:
@astronaut-wannabe what I can't seem to crack is if the shared-db thread thing applies to all Capybara stacks or only Rails 5.1's SystemTests

astronaut-wannabe:
@maschwenk It only applies to rails 5.1+ (the changes went into rails repo, not capybara repo)

Here is the PR that added the shared-db thread stuff: https://github.com/rails/rails/pull/28083

from reading that, it applies to all rails 5.1 tests that use multiple threads.

it just-so-happens that the reason you always needed to use database cleaner prior to 5.1 was that:
> When a system test starts Puma spins up one thread and Capybara spins up
another thread. Because of this when tests are run the database cannot
see what was inserted into the database on teardown. This is because
there are two threads using two different connections.

So, now in rails 5.1+ all threads in a test (any test (***I think***) not just system test) will be able to see the same DB.

capybara + system_test just seems to be the most common use case.

maschwenk:
@astronaut-wannabe right, but when not using SystemTests, Capybara defaults to using Webrick, which it is unclear whether or not this feature would affect Webrick driven SystemTests (for example using a more standard `rspec-rails` setup)

veganstraightedge:
@guitsaru Look good to you?

guitsaru:
@veganstraightedge yep, looks good to me.

Issue #394, opened 2017-07-30
Add 'Subscribe' to header nav


veganstraightedge:
- [ ] **on homepage:** link to #anchor to footer signup form
- [ ] **all other pages:** link to homepage + #anchor to footer signup form
- [ ] do the right thing with nav on different viewport sizes (maybe just hide it when the nav goes grid?)

PR #402, opened 2017-08-06
#394 Add the subscribtion menu to the header

just1602:
@veganstraightedge, it could be cool if you had some time to review the css and tell me if it's ok or if you sea a better way to do it. :)

### DIFF
--- a/app/assets/stylesheets/components/_navigation.scss
+++ b/app/assets/stylesheets/components/_navigation.scss
@@ -30,7 +30,6 @@
   }
 }
 
-
 .secondary-navigation {
   display: flex;
   align-items: center;
@@ -67,7 +66,6 @@
   }
 }
 
-
 // small viewports
 .site-header nav[role=navigation],
 .secondary-navigation {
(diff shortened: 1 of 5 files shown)


veganstraightedge:
Merged. I have additional removed /video from the nav for now to keep it to two rows of nav items.

just1602:
But, Subscribe is to long for the design. When the width get >= 800px, than the menu get horizontal, but with the subscribe item, it's too long and the bulb logo goes under the Crimethinc. title. It takes like 810px for the bulb logo to take the right place.

Do you think we could change the $small-phone viewport width or just find a way to make the nav smaller ?

veganstraightedge:
Really there should be a responsive menu with a hamburger icon. :/ 

Is that something you'd be up for doing if I sketch up some comps?

just1602:
Yeah, if you have an idea on how you want it I could do it.

veganstraightedge:
Great. I'll sketch it later today and create a new issue for you.

Issue #399, opened 2017-08-03
Search error : with quote in query


veganstraightedge:
`https://crimethinc.com/search?utf8=%E2%9C%93&q=test+their+logik+democracy%27s+bankrupt`

***

```
vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:620:in `async_exec': PG::SyntaxError: ERROR:  syntax error at or near "s"
LINE 1: ...ent, phraseto_tsquery('test their logik democracy's bankrupt...

: SELECT  search_results.*, ts_rank(document, phraseto_tsquery('test their logik democracy's bankrupt')) AS ranking FROM "search_results" WHERE (document @@ phraseto_tsquery('test their logik democracy''s bankrupt')) ORDER BY ranking DESC LIMIT $1 OFFSET $2 (ActionView::Template::Error)
    from vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:620:in `block (2 levels) in exec_no_cache'



Caused: vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:620:in `async_exec': PG::SyntaxError: ERROR:  syntax error at or near "s"
LINE 1: ...ent, phraseto_tsquery('test their logik democracy's bankrupt...
                                                             ^
: SELECT  search_results.*, ts_rank(document, phraseto_tsquery('test their logik democracy's bankrupt')) AS ranking FROM "search_results" WHERE (document @@ phraseto_tsquery('test their logik democracy''s bankrupt')) ORDER BY ranking DESC LIMIT $1 OFFSET $2 (ActiveRecord::StatementInvalid)
    from vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:620:in `block (2 levels) in exec_no_cache'



Caused: vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:620:in `async_exec': ERROR:  syntax error at or near "s"
LINE 1: ...ent, phraseto_tsquery('test their logik democracy's bankrupt...
                                                             ^
 (PG::SyntaxError)
[...]

PR #404, opened 2017-08-07
[#399] Search error : with quote in query

astronaut-wannabe:
resolves https://github.com/crimethinc/website/issues/399

---

The search feature uses PostgreSQL's `tsquery` feature to enable
searching.

The [documentation indicates that][0]

> PostgreSQL provides the functions to_tsquery and plainto_tsquery for converting a query to the tsquery data type. to_tsquery offers access to more features than plainto_tsquery, **but is less forgiving about its input**

The actual issue is that each term passed to `phraseto_tsquery` must
be surrounded by single quotes.

There seems to be 3 options for making sure single-quotes in user searches don't cause
syntax errors:

1. using `plainto_tsquery` instead (doesn't seem like a good search experience)
2. passing query to [`quote_literal`][1] twice first (didn't seem to work when I tried it)
3. just stripping single quotes in the controller (what I ended up doing)

---

**Note on advanced search:**
Advanced search had this issue too, but fixing simple-search also
fixes advanced. This is because advanced search just parses the form
entry and then redirects to simple-search with the parsed form data as
the query string (so the search ultimately ends up flowing through the
`search_params` method I added.

[0]:https://www.postgresql.org/docs/9.3/static/textsearch-controls.html
[1]:https://dba.stackexchange.com/questions/135030/filtering-special-characters-in-to-tsquery

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -7,7 +7,7 @@ class SearchController < ApplicationController
       return redirect_to root_path
     end
 
-    @search  = Search.new(params[:q])
+    @search  = Search.new(search_params)
     @results = @search.perform.page(params[:page]).per(15)
     @title = "Search results for \"#{@search.query}\""
   end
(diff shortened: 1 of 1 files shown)


astronaut-wannabe:
@veganstraightedge I have never worked with PostgreSQL before, and it *seems* like there should be a database-level way to handle this...

but maybe not, I wasn't able to really find anything other than making sure single-quotes are not passed to Postgres

PR #433, opened 2017-08-14
Fix slug_spec

astronaut-wannabe:
The reason this spec was failing was because you cannot directly do a
`Page.create!(slug: "test")` and get a page created with a slug of
`"test"`.

Not sure why as I haven't reall used Rails `Concerns` much, but
setting the `:title` to `"test"` will give that `Page` a slug of
`"test"`, allowing us to test the "when a slug already exists" case
correctly.


astronaut-wannabe:
this PR combined with #434 will make that ❌ a ✅

PR #434, opened 2017-08-14
fix categories_controller_spec

astronaut-wannabe:
in d11578aa7354d4970d812efb8fcd8620089324b2 a redirect was added to
the `#show` action in `categories_controller.rb`.

So, this spec is actually correct. We just need to updated the
expectation to reflect the added redirect


astronaut-wannabe:
this PR combined with #433 will make that travis ❌ a ✅

Issue #450, opened 2017-08-25
Get Heroku CI to play nice with headless chromedriver


veganstraightedge:
Right now, Travis CI passes, but Heroku CI fails. Heroku fails because it doesn't know what to do about that headless chromedriver. Can you see if you can get this fixed up on a test Heroku app, @astronaut-wannabe so that our build goes back to green in Heroku CI land, please?

```
-----> Running Ruby buildpack tests...
-----> Running test: bin/rspec
       .....................................................F......................................................................................................................................
       Failures:
       1) Setting and changing an articles published_at date Using 'PUBLISH NOW' feature
       Failure/Error: visit '/signin'
       Selenium::WebDriver::Error::WebDriverError:
       Unable to find chromedriver. Please download the server from http://chromedriver.storage.googleapis.com/index.html and place it somewhere on your PATH. More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver.
       # ./spec/spec_helper.rb:25:in `login_user'
       # ./spec/features/article_datetime_settings_spec.rb:81:in `block (2 levels) in <top (required)>'
       Finished in 31.91 seconds (files took 5.17 seconds to load)
       188 examples, 1 failure
       Failed examples:
       rspec ./spec/features/article_datetime_settings_spec.rb:75 # Setting and changing an articles published_at date Using 'PUBLISH NOW' feature
       Coverage report generated for RSpec to /app/coverage. 5411 / 12970 LOC (41.72%) covered.
-----> Ruby buildpack tests failed with exit status 1
```

astronaut-wannabe:
sure thing, will fix this tomorrow

veganstraightedge:
Thanks! 

Also, I've invited you to be a member of the repo, so that issues can be assigned to you proper and PRs can be branches on this repo.

astronaut-wannabe:
👍

astronaut-wannabe:
resolved in #452

PR #452, opened 2017-08-25
[#450] Get Heroku CI to play nice with headless chromedriver

astronaut-wannabe:
This change adds the needed build-packs mentioned in the [Heroku Docs][0]
to get chrome headless to run.

That said, there might be alternative changes we could make:

- since specs already ran in Travis, just skip them in Heroku
- since feature specs take the longest (and already ran in travis), skip the feature specs in heroku
- since specs will run in heroku, skip the specs in travis
  - I am not a fan of this option as setting up heroku on a repo fork is harder than setting up travis

It just seems redundant (not in a useful way) to run the specs multiple times.

[0]:https://devcenter.heroku.com/articles/heroku-ci-browser-and-user-acceptance-testing-uat#google-chrome-via-headless

### DIFF
--- a/app.json
+++ b/app.json
@@ -34,5 +34,15 @@
     "heroku-postgresql:hobby-dev"
   ],
 
-  "image": "heroku/cedar"
+  "image": "heroku/cedar",
+
+  "environments": {
+    "test": {
+      "buildpacks": [
+        { "url": "heroku/ruby" },
+        { "url": "https://github.com/heroku/heroku-buildpack-chromedriver" },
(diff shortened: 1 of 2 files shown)


veganstraightedge:
I see your point. And generally agree with you. But where we can, we also like to find redundancy so that if one thing fails, we have a backup plan. (Like NASA!)

If this was very hard to do, I'd say skip tests on Heroku CI. But since you've already done it, I say we ship it and continue to CI in both places.

Issue #454, opened 2017-08-26
FIX: Intermittent spec failure


veganstraightedge:
# First run, all pass

```
> rspec
............................................................................................................................................................................................

Finished in 12.18 seconds (files took 2.8 seconds to load)
188 examples, 0 failures
```

# Second run, several fail

```
> rspec
WARN: Unresolved specs during Gem::Specification.reset:
      diff-lcs (< 2.0, >= 1.2.0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
...................................................FFF....................................................................................................................FFFFFFFFFF..F.....

Failures:

  1) Setting and changing an articles published_at date updating an existing article
     Failure/Error: expect(page).to have_content 'Logged in!'
       expected to find text "Logged in!" in "Retry later"
     # ./spec/spec_helper.rb:31:in `login_user'
     # ./spec/features/article_datetime_settings_spec.rb:38:in `block (2 levels) in <top (required)>'

  2) Setting and changing an articles published_at date Saving an article without entering publication date info
     Failure/Error:
       within('main') do
         fill_in 'username', with: 'user1'
         fill_in 'password', with: 'c'*31
       end

     Capybara::ElementNotFound:
       Unable to find visible css "main"
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/node/finders.rb:313:in `block in synced_resolve'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/node/base.rb:85:in `synchronize'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/node/finders.rb:302:in `synced_resolve'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/node/finders.rb:37:in `find'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/session.rb:776:in `block (2 levels) in <class:Session>'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/session.rb:333:in `within'
     # /Users/s/.gem/ruby/2.4.1/gems/capybara-2.15.1/lib/capybara/dsl.rb:50:in `block (2 levels) in <module:DSL>'
[...]

PR #455, opened 2017-08-27
[#454] disable rack-attack if Rails.env.test?

astronaut-wannabe:
Rack attack was causing HTTP response `429` if you ran the test suite
back-to-back several times.

tested the fix by running:
```sh
while true; do rspec; done
```

and watched it for several runs

resolves: https://github.com/crimethinc/website/issues/454


astronaut-wannabe:
err, I don't know what best practice is. maybe it would be better to leave the initializer untouched and just move
```ruby
 config.middleware.use Rack::Attack
```
from [`/config/application.rb`][0] to `/config/environments/production.rb`

[0]:https://github.com/crimethinc/website/blob/c18106d00026cf6aaaa6b0fb1ac66118108fcf1d/config/application.rb#L17

Issue #416, opened 2017-08-11
Add published_at / draft mode to Tools


veganstraightedge:
- [ ] Only show published ones on /tools pages, like Article etc

greggawatt:
id like to start on this one this week, Anything I need to know about it?

veganstraightedge:
Look to `Article` for model attributes, scopes and UI. `Status` and `published_at` are the big ones.

greggawatt:
just to clarify, by tools you mean the logo, sticker, zine, poster, and video models right?

veganstraightedge:
yes and...

- books and zines both use the Book model
- stickers and posters both use the Poster model
- logos are Logo
- videos are Video

greggawatt:
I'm thinking for this a parent model might make sense with all the published logic in it so it's consistent across all models. If not that, maybe a shared library? Kinda curious about architecture thoughts here.

veganstraightedge:
it feels like a situation where a Concern could be useful.

See what @guitsaru has already done with concerns so far.

https://github.com/crimethinc/website/tree/master/app/models/concerns

greggawatt:
I'm making a concern called publishable to take care of this. I see the following in post concern: 
  def published?
    status.name == "published"
  end

I was considering saying
  def published?
    published_at.present?
  end

But it seems like we should be consistent, Should all the tools have status_id as well and we treat everything like a state machine?

veganstraightedge:
Closed by https://github.com/crimethinc/website/pull/483

PR #483, opened 2017-09-17
/tools indexes now only show published content

### LINKED ISSUE #481: Add Slug concern to Video
Seems like this should have `Slug` like all the rest of the models that have `Name` and `Publishable`

greggawatt:
fixes #481 and #416


greggawatt:
@veganstraightedge This is the finished solution to the issue. Any feedback on how i chose to do it?

PR #488, opened 2017-09-28
Partial for issue 451

explody:
Does work for pinging archive.org, does not include related automation. Manual tests only since it hits archive.org.


lib/tasks/wayback.rake
```diff
@@ -0,0 +1,35 @@
...
+    request = Net::HTTP::Post.new(uri)
+    request.content_type = 'application/json'
+    request.body = JSON.dump('url' => url)
+
+    response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
+      http.request(request)
+    end
+    if response.code == '200'
+      r = JSON.parse(response.body)
+      puts "#{url} => https://web.archive.org#{r['wayback_id']}"
```
astronaut-wannabe: probably change these `puts` to `info` logs (or does that happen automatically?)

lib/tasks/wayback.rake
```diff
@@ -0,0 +1,35 @@
...
+    urls.append args[:url]
+  else
+    since = 1.day.ago.to_s
+    Article.where("updated_at > '#{since}'").each do |article|
+      urls.append [app['website'], article.path].join('')
+    end
+  end
+
+  uri = URI.parse(api)
+  urls.each do |url|
```
astronaut-wannabe: is there a rate limit?

I guess there won't be _that_ many urls, so maybe no need to worry about an accidental ddos

lib/tasks/wayback.rake
```diff
@@ -0,0 +1,35 @@
+desc 'Request the Wayback Machine to snapshot a specific URL or all new '\
+  'article URLs for the past 24 hours if no argument is given'
+
+task :wayback, [:url] => :environment do |_t, args|
+  app = JSON.parse(File.read(Rails.root.join('app.json')))
```
astronaut-wannabe: i think it is fine to hardcode the website here, can't really think of a situation where it would change

lib/tasks/wayback.rake
veganstraightedge: I think that happens automatically.

lib/tasks/wayback.rake
veganstraightedge: I have a rate limit before but I was blasting through every single article on the site. I can't imagine that we'll run into it with just today's articles.

lib/tasks/wayback.rake
astronaut-wannabe: 👍

lib/tasks/wayback.rake
veganstraightedge: Either way is fine by me. I'ma merge this as is. @explody if you want to hardcode this in another PR, cool.

PR #489, opened 2017-09-28
[WIP] [Feature suggestion] Check passwords on user sign-up against DB of known leaks

astronaut-wannabe:
I was going through a bunch of accounts on [HaveIBeenPwned][0] and I
noticed the [Passwords][1] tab for the first time.

That led me to reading this [post][2] about checking against this
password DB.

This PR specifically implements the registration use case [mentioned here][3].

There are still several technical issue to resolve, but I wanted to
open a PR to start getting feedback before continuing.

My open questions are:

1. Do we want to do this?
2. Would we want to implement the other Use Cases mentioned?
3. If so, what is the best way to store this giant password DB (It's 5.7 GB *compressed*)
4. A faster way to do the DB lookup (probably don't want to use ActiveRecord for this)

[0]:https://haveibeenpwned.com/
[1]:https://haveibeenpwned.com/Passwords
[2]:https://www.troyhunt.com/introducing-306-million-freely-downloadable-pwned-passwords/
[3]:https://www.troyhunt.com/introducing-306-million-freely-downloadable-pwned-passwords/#usecase1registration

### DIFF
--- a/app/models/burned_password.rb
+++ b/app/models/burned_password.rb
@@ -1,0 +1,3 @@
+class BurnedPassword < ActiveRecord::Base
+  self.primary_key = 'password_sha1'
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,6 +15,7 @@ class User < ActiveRecord::Base
               in: ["mickey fickie fire cracker soap on a rope", "a long passphrase to meet the minimum length"],
               message: "The passphrase '%{value}' is prohibited."
             }
+  validate :password_not_already_burned, on: :create
 
   default_scope { order(username: :asc) }
 
@@ -27,5 +28,12 @@ class User < ActiveRecord::Base
 
   def name
     username
+  end
+
+  private
+
+  def password_not_already_burned
+    burned = BurnedPassword.exists?(Digest::SHA1.hexdigest(password))
+    errors.add(:password, :burned_password) if burned
   end
 end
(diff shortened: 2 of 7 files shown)


veganstraightedge:
I like this idea in theory. But in practice, fuuuuuck, that's a big DB.

Does that site provide a way to check individual passwords without having to commit to hosting that giant thing ourselves?

My pragmatic concern is that 5.7gb of database records will push us up into a much higher tier of database pricing. And we're a tiny volunteer operation. 

Alternate idea, what's the level of effort to implement two factor auth? We'd get the same security benefit without the baggage of massive db increase. Even if someone is using "password" as their password, if they have to 2FA with an authy-like app, then their account is safer. Right?

astronaut-wannabe:
> Does that site provide a way to check individual passwords without having to commit to hosting that giant thing ourselves?

yes, but it says pretty explicitly the API is just for playing around/testing. We don't want to start sending passwords over the wire and to 3rd parties (even trusted-ish ones)

> My pragmatic concern is that 5.7gb of database records will push us up into a much higher tier of database pricing. And we're a tiny volunteer operation.

agreed, I also figured this would be a major hurdle

> Alternate idea, what's the level of effort to implement two factor auth? We'd get the same security benefit without the baggage of massive db increase. Even if someone is using "password" as their password, if they have to 2FA with an authy-like app, then their account is safer. Right?

I have just started reading up on what it takes to implement authy stuff. will get back to you

veganstraightedge:
@astronaut-wannabe Let's say we close this pull request? And open a new issue for 2FA if you want to go that route?

PR #509, opened 2017-10-25
Adds library, tools, and yearly archive directories to sitemap

### LINKED ISSUE #502: Update sitemap.xml/.rb
We've added/changed some paths. They need to make it into the sitemap.

/library
/podcast
/tools

Maybe more. It's worth exploring the site to make sure we're covered.

gesa:
closes #502 

Here's a thing: the sitemap should be updated at build time by running `rake sitemap:refresh`. That not only updates the xml, but also pings search engines.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -33,11 +33,13 @@ SitemapGenerator::Sitemap.create(default_host: "https://crimethinc.com", compres
       "/get/",
       "/kickstarter/2017/",
       "/listen/",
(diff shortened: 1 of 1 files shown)


veganstraightedge:
@gesa Should `rake sitemap:refresh` be run on a local machine then checked in or on production?

gesa:
On production. On local/staging environments, run `rake sitemap:refresh:no_ping`

PR #537, opened 2017-11-29
add tags to books, zines, posters, stickers

photomattmills:
Hopefully this is right, I just cargo-culted the setup from articles and wrote a couple basic tests. Issue #390  [here](https://github.com/crimethinc/website/issues/390).


app/models/book.rb
```diff
@@ -4,6 +4,11 @@ class Book < ApplicationRecord
   scope :book, -> { where(zine: false) }
   scope :zine, -> { where(zine: true)  }
 
+  has_many :taggings, dependent: :destroy, as: :taggable
```
veganstraightedge: All of this looks good. 👍🏻

Small nit pick though, if you don't mind, will you reorder this so `dependent: :destroy` is last please?

I just think it reads better when it's:

`has_many :things, as: :thing_able, …`

Issue #176, opened 2017-01-01
Content for Error pages


veganstraightedge:
# 404
## File not found

- [ ] Sub/Title
- [ ] Image
- [ ] Paragraph or two
- [ ] Suggested paths from here (search, /read, etc)

# 500
## Internal App Error (we fucked up)

- [ ] Sub/Title
- [ ] Image
- [ ] Paragraph or two
- [ ] Suggested paths from here (search, /read, etc)

biofilo:
Error 404: File Not Found

http://lapiedradesisifo.com/wp-content/uploads/2015/08/Biblioteca.jpg <http://lapiedradesisifo.com/wp-content/uploads/2015/08/Biblioteca.jpg>
or
http://dorothygebert.com/wp-content/uploads/2014/04/enhanced-buzz-wide-1782-1374872169-42.jpg <http://dorothygebert.com/wp-content/uploads/2014/04/enhanced-buzz-wide-1782-1374872169-42.jpg>

The page you are looking for does not seem to be here. To go hunting for it, [use our sophisticated search engine,](LINK?) or peruse our [reading library.](LINK) To report a problem to us, [DO THIS.]


Error 500: Internal App Error

http://static.travelinglifestyle.net/uploads/2016/04/travel-inspiration.jpeg <http://static.travelinglifestyle.net/uploads/2016/04/travel-inspiration.jpeg>

The page you are looking for does not seem to be here. To go hunting for it, [use our sophisticated search engine,](LINK?) or peruse our [reading library.](LINK) To report a problem to us, [DO THIS.]

Issue #157, opened 2016-12-29
Make custom 404, 500 and 422 pages


veganstraightedge:
Content: https://github.com/crimethinc/website/issues/176

veganstraightedge:
- [ ] *404* : https://github.com/crimethinc/website/blob/master/public/404.html
- [ ] *500* : https://github.com/crimethinc/website/blob/master/public/500.html

veganstraightedge:
@mousemke Can you take a first pass at this? Those pages have to be completely self sufficient. No SCSS or external CSS. I'll upload the images to S3.

veganstraightedge:
Images:

http://thecloud.crimethinc.com/assets/errors/404.jpg
http://thecloud.crimethinc.com/assets/errors/500.jpg

mousemke:
are we sure having [1mb and 2mb respectively] images on the 404 and 500 pages is appropriate? Seems kind of unnecessary. 

Also, I tried loading the 500 image on 3g and it took 22 seconds.

![http://pics.knoblau.ch/500.jpg_26672000_2017-01-02_20-42-41.png](http://pics.knoblau.ch/500.jpg_26672000_2017-01-02_20-42-41.png)

veganstraightedge:
You can downsize it as necessary. When you get something good, email it to me and I'll reupload.

mousemke:
cool

just1602:
@astronaut-wannabe do you think we should also do a 422 page ?

astronaut-wannabe:
@just1602 not sure. It would be simple enough, but I think that would only ever really happen on `admin` pages, so maybe not needed

just1602:
@astronaut-wannabe, nice that means we can close this issue ! :D

veganstraightedge:
Agreed. We don't need a 422. 

Thanks for doing this one. 👍🏻

PR #546, opened 2017-12-04
Add a custom 500 page [#157]

astronaut-wannabe:
- I added the image linked to in [issue #176][0]
- created different sized versions of image
- added the text from the above issue

TODO
----------
- [x] design guidance on text and link styles
- [x] the copy says "page not here", which is sort of wrong for a `500`
- [x] work on the mobile-sized/cropped breakpoints a little more (newer smartphones can probably have a non-cropped version due to hi dpi)
- [x] match site's `p` `max-width`

Demo
---------------
https://crimethinc-staging-pr-546.herokuapp.com/500.html

[0]:https://github.com/crimethinc/website/issues/176


veganstraightedge:
Very exciting! Thanks for picking this one up. It's been in the back of my brain for a long time.

veganstraightedge:
>  the copy says "page not here", which is sort of wrong for a 500

How about something like:

*** 

# Error 500: Internal App Error

**Something went wrong.** It's probably our fault, not yours. (We all alerted of all errors on the site and will try to fix whatever is broken.) 

To go hunting for the page you want, [use our sophisticated search engine](/search/advanced), or peruse our [reading library.](/library) To report a problem to us, [email us](mailto:tech@crimethinc.com).

veganstraightedge:
>  design guidance on text and link styles

I would make the links the same color as the rest of the text. Underlined is good. Maybe embolden them too.

Color: I think white text could look really nice. It might require dropping the white point to `0.9` (in the Levels palette) on that image.

Also try putting the non-heading text in a black box with some comfortable padding and a `.5`ish transparency on the background black.

public/500.html
```diff
@@ -1,66 +1,53 @@
...
+      <source media="(min-width: 36em)"
+              srcset="500_assets/full_size.jpg  2667w,
+                      500_assets/large.jpg  1024w,
+                      500_assets/medium.jpg 640w,
+                      500_assets/small.jpg  320w"
+              sizes="100vw" />
+      <source srcset="500_assets/cropped-large.jpg 2x,
+                      500_assets/cropped-medium.jpg 1.5x,
+                      500_assets/cropped-small.jpg 1x" />
+      <img src="500_assets/small.jpg" />
```
veganstraightedge: I think I'd prefer if there was a directory called `errors` or `error-pages` with `500` and `404` directories inside of it.

veganstraightedge:
Nav.

I know this isn't in the issue / spec, but I wouldn't hate if these error pages had the header/logo/nav on them too. If you feel like that's out of scope for this issue, that's cool. We can make it into its own issue if you prefer.

public/500.html
astronaut-wannabe: sounds good.

just a note, 404 will never be seen (we have that global re-direct to the homepage)

veganstraightedge: Good point!

Though, I think it's maybe better to land on a 404 (and report it in the logs) than just opaquely redirect to root. I think I initially made that root redirect when things were still messy and in flux at the start. Getting rid of it might help identify where things ARE actually missing or broken.

What do you think about removing it and adding a static 404?

astronaut-wannabe: @veganstraightedge agreed. I will drop the redirect as part of the 404 PR

veganstraightedge:
Progress is looking better. Here's some more feedback:

- [ ] Change "Error 500: Internal App Error" to "500: Internal App Error"
- [ ] Add a wrapper element around the two `<p>`s and give that the black background, so that they're in the same box
- [ ] actually try putting the heading in that box too. see how that looks.

Looking good!

astronaut-wannabe:
@veganstraightedge 

In these 4 screenshots the `h1` is either `2em` or `3em` and included in the background or not. Which do you prefer?

## `2em` for `h1`
<details><summary>header in background</summary>

2em-header-with-back

</details>
<details><summary>header above background</summary>

2em-header-no-back

</details>

## `3em` for `h1`
<details><summary>header in background</summary>

3em-header-with-back

</details>
<details><summary>header above background</summary>

3em-header-no-back

</details>

veganstraightedge:
I like `2em for h1 + header in background` 👍🏻

astronaut-wannabe:
@veganstraightedge I think this is done enough for me to move onto the 404.html. after that I will add the site header to both

not sure if the images need to be uploaded to the cdn instead of stored in `/public`, but I wanted to see how it performed in prod to see if I needed to cut any more image variations

veganstraightedge:
Great work on this, @astronaut-wannabe! ‼️⭐️🌟🎉

I agree that doing 404 as a separate PR and then adding nav to both is a good idea.

I'll move the image to S3/CloudFront and update the URLs in the `` tags.

PR #564, opened 2017-12-10
Add custom 404

astronaut-wannabe:
- `cp 500.html 404.html`
- make the text match what was requested in issue #157
- add 404 image assets (upload to cdn later)

### demo
https://crimethinc-staging-pr-564.herokuapp.com/404.html


app/controllers/posters_controller.rb
```diff
@@ -19,7 +19,7 @@ def show
     if @product.present?
       @product = @product.first
     else
-      return redirect_to [:posters]
```
just1602: Do you think we can use a `find_by!` instead of a `where` call for the product controllers and then delete the if statement that raise the not found? I'd simplify the logic like in the others controllers.

veganstraightedge: I'm not opposed to it. Will it just error out if there's no record with that id/slug? The reason I did the `where` + `if` is to do a `redirect` to a place closer to what they were trying to find instead of just a 500 page.

veganstraightedge: @just1602 If you think this is something you want to explore, will you make a new issue so we can have this conversation there, please?

just1602: I totally agree to not throw a 500, but if `find_by!` throw a 404 when it didn't found the queried model, it sounds good to me. It would be the usual use case, instead, when you redirect to the nearest resources like `:posters` here, the user doesn't necessary know if it's because the resource doesn't exists or it's just a bad url.

But I'm a big fan of clean code, TDD and other stuff like that, so maybe I just give to much attention to those details. :P

astronaut-wannabe: @veganstraightedge @just1602 there are a group of errors rails will rescue and direct to the 404 page. `RecordNotFound`is one. All of the changes I made will cause a 404, not a 500.

I changed most things to `find_by!`, but left the `where` and the `if`/`else` in cases where I wasn't 100% sure of what the code was doing or if it seemed like a re-direct was more useful.

For example, I left the redirect in categories bc  it will take you to the index of all categories (which is nice).

veganstraightedge: @just1602 Sorry. When I suggested a new issue for discussing this, I didn't realize that this PR was doing more than just 404.html styles. You were right to discuss the `where`/`if/else`/`redirects` here. My bad. I'll close that issue to continue the discussion here.

veganstraightedge: @astronaut-wannabe I have mixed feelings about this. I think raising 404 makes sense in most places where things aren't found. _Buuuuut_ I feel like what is more useful to readers is redirect the collection of that type of thing if there is a collection page. So, what do you think about this?

1. Find the resource that is requested in the URL.
2. If it's not there **and** there's a collection page, `redirect_to` to the collection page and add an alert/notification explaining what happened. 
3. If there's not a collection page, raise 404.

The alert copy could be something similar to the 404 text, but dynamic and a little more context.

> You requested `/books/xyz` which doesn't exist. This page contains all of our books. If you can find it here, you can try using our `search engine` or perusing the `library`.

Or whatever.

PS. Sorry for not thinking this through before you got started on this. I hadn't considered more than "just" an overhaul of the HTML/CSS of 404.html. But what you're doing makes sense. Good looking out. 👍🏻

veganstraightedge: Pasting this here for posterity. 

@just1602's description on issue #565.

> I was wondering if we should redirect people to the appropriate resources (exemples, redirecto to /categories, when a category isn't fonud) or should we throw a 404 error?
> 
> Personally, I prefer the 404 error, because it is the usual workflow on the web. I also think it's easier to know for a user that he or she has the wrong url, rather than just be redirect to another page. Imagine if you're a blind person using a reading machine you try to access an url which, for any reason has a typo, and you're redirect to the better resource for this url, you may ask you some question about what's going on. The other thing is no 404 error should be thrown, otherwise than when someone hand type an url and make a typo, for the rest it's broken links and it could be nice to have notification for them.
> 
> @veganstraightedge @astronaut-wannabe and the others, what do you think about it? :)

just1602: If an flash message or something tell the user that the resource requested doesn't exists and he has been redirect to the collection page, I'm fine with that. I only see a problem with the redirect when you're redirect elsewhere you wanna go and you don't know why.

astronaut-wannabe: @veganstraightedge @just1602 

Ok, I will remove the logic changes for this PR and put them in a new one so it is easier to review.

veganstraightedge: @astronaut-wannabe How do you feel about a notification + redirect?

astronaut-wannabe: @veganstraightedge sounds good to me

astronaut-wannabe:
@veganstraightedge @just1602 this PR now *only* includes the `404` asset changes

PR #572, opened 2017-12-14
 [TCE] Make jquery assets load from http

astronaut-wannabe:
The non-english version of TCE are not loading their
images.

Pointing jquery at `https` fixes that


astronaut-wannabe:
@veganstraightedge  So, this change *does* bring back images [in staging][0]. It is weird though, because the script assets **are** in the cdn. Maybe the cdn version is different from the local version

[0]:https://crimethinc-staging-pr-572.herokuapp.com/tce/espanol

PR #573, opened 2017-12-14
[TCE] point tce espanol get page at right version

astronaut-wannabe:
The 'download pdf' link on the espanol page takes you
to the english `get` page.


astronaut-wannabe:
not sure why relative paths aren't working, so I went absolute

Issue #80, opened 2016-12-05
Build Share sheet


veganstraightedge:
View on mobile:
http://www.crimethinc.com/texts/r/battle/

Click the little share icon that kinda like a `<`.

Build something similar to that for:
- Articles
- Pages

When not on a Page or Article, fallback to sharing the Homepage.

***

I'd really prefer that this is built without Javascript. I'm thinking that the `:target` pseudo selector can be used to hide/show the share sheet, since we don't have to worry about scrolling since we're already at the top of the window.

***

```html
<a href="#share">SHARE</a>

<div id="share" class="hidden">
  [share sheet content]
</div>
```

```css
:target {
  display: block;
}
```

Apologies if you already knew this technique.

veganstraightedge:
**New idea.**

Redesign the Article footer to include share links to: twitter, facebook, tumblr.

Instead of a share sheet in the header/nav.

PR #576, opened 2017-12-14
Add the share sheet #80

just1602:


just1602:
Sounds good, can be merge !

just1602:
Sounds good, can be merge !

PR #577, opened 2017-12-15
Add translation yaml plumbing for TCE

astronaut-wannabe:
## Overview
I have been adding in some plumbing to make it easier to create web versions of the rest of the TCE languages. I then localized `tce/espanol` using the new template at `tce/es` so that you could see and example. An overview of the structure (using es as the example) is this:

- The `<head>` of  [website/public/tce/index.html][0] was pulled out into `app/views/layouts/to_change_anything.html.erb`
- the page content was pulled out into `es.yml`
- language specific css was moved to `assets` and wrapped in a `.es {}` scss block
  - `class="es"` is then applied to the `<body>` so that language-specific css can still be supported

## TODO
- [x] translate Table of Contents menu for `/es`
- [x] translate Share menu for `/es`
- [x] figure out how to handle dialects (e.g. ES vs ES-latin-america)
  - [w3.org langvalues][1]
  - [iana.org language-subtag-registry][2]
  - [rfc bcp47][3]
- [x] add `ltr` and `rtl` language toggles
- [x] ~cache pages since they basically never change~ (will do in different PR)

## demo
you can compare
 - https://crimethinc-staging-pr-577.herokuapp.com/tce/es          (new method)
 - https://crimethinc-staging-pr-577.herokuapp.com/tce/espanol (original)

### Image Diff
I have been performing image diffs between `/tce/espanol` and `/tce/es` while I developed.
```sh
diffimg tce-espanol.png tce-es.png > diff.png
```
The latest page diff is a satisfying solid black square (e.g. no diff)

<details><summary>see page diff image</summary>

diff

</details>

[0]:https://github.com/crimethinc/website/blob/master/public/tce/index.html
[1]:https://www.w3.org/International/questions/qa-html-language-declarations#langvalues
[2]:http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
[3]:http://www.rfc-editor.org/rfc/bcp/bcp47.txt


veganstraightedge:
> create redirect first?

Please do. 

You can either do that in `redirects.txt`:

```
from_path to_path
```

Or you can tell me what you want and I can create it in the production database.

app/controllers/to_change_everything_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  RTL_LANGUAGES = [
+    :ar, # Arabic
+    :ku, # Kurdish
+    :he, # Hebrew
+    :fa, # Persian
+  ].freeze
+
+  def show
+    http_cache_forever(public: true) do
```
astronaut-wannabe: I am a little nervous about using `http_cache_forever`. It seems like the exact thing I want, but I don't think there is a way to bust the cache besides people manually hard-refreshing their browsers...

astronaut-wannabe: err, maybe this isn't what I want. Seems like every new visitor will get 1-slow request.

I want 1 visitor to get 1 slow request, and then everyone to get the cached view afterwards

just1602: Maybe we can use the usual caching system, but set a very long timeout for those pages.

app/views/layouts/to_change_everything.html.erb
```diff
@@ -0,0 +1,263 @@
+<% cache @locale, expires_in: 1.year do %>
```
astronaut-wannabe: this gives a huge performance boost when I test locally, but I see no difference in staging...

astronaut-wannabe:
can someone sanity-check this font asset url change: https://github.com/crimethinc/website/pull/577/commits/bfb6b165c9a01daf7a0a89cf2c8d072e03f7de7f

astronaut-wannabe:
@veganstraightedge what do you think about this plan.

- remove caching for now
- deploy `/es` to prod and have it run side-by-side with `/espanol` to see if any issues are discovered before adding a re-direct
- if everything looks good, add caching
- if everything looks good after that, add redirect

The main reason I am suggesting this is because github diffs are not very useful right now for distinguishing between what is a straight copy-paste from `/public/tce` and what is a change I am making

veganstraightedge:
I think launching without caching is a good idea.

veganstraightedge:
I'm unclear about the URL change from `espanol` to `es`. 

- Is that just to make the URL match the lang yaml filename? 
- Is the plan to change all of the languages URLs to two letter codes?
- What's the degree of difficulty to keep them as their language names instead of codes?

astronaut-wannabe:
> I'm unclear about the URL change from espanol to es.

just so I could run the original version and the template versions side-by-side for comparisons while developing.

> Is that just to make the URL match the lang yaml filename?

yes. I think I can name the yaml files whatever, but I was just following http language-region convention 

> Is the plan to change all of the languages URLs to two letter codes?

Most of the languages do not have urls at all, so the plan was to have all the new ones be language codes.

> What's the degree of difficulty to keep them as their language names instead of codes?

it should be simple

veganstraightedge:
Hmm. I guess there's also the question of how to handle language names that are in non-URL safe characters.

veganstraightedge:
> > I'm unclear about the URL change from espanol to es.

> just so I could run the original version and the template versions side-by-side for comparisons while developing.

Oh duh. That's clever. 👍🏻 @astronaut-wannabe

veganstraightedge:
@astronaut-wannabe Would you mind stripping all of the wrapping quotation marks that are unnecessary in the es.yml? I think it's cleaner without them.

astronaut-wannabe:
@veganstraightedge, will do.

It will be very mixed though. A lot of the fields have internal `:`s or alpha-numeric characters such as `&nbsp;`

app/views/layouts/to_change_everything.html.erb
```diff
@@ -0,0 +1,261 @@
+<!DOCTYPE html>
+<html dir="<%=t('.language_direction')%>" lang="<%=t('.lang_code')%>" prefix="og: http://ogp.me/ns#" <%= html_id %>>
```
astronaut-wannabe: I think I cargo culted `prefix="og: http://ogp.me/ns#" <%= html_id %>` from somewhere

I don't know what it is or if I need it

just1602: open graph is meta tags use by Facebook, Tumblr and others websites to get information about an app or a page you share.

app/controllers/to_change_everything_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ToChangeEverythingController < ApplicationController
+  layout 'to_change_everything', only: [:show]
+  skip_before_action :set_pinned_pages, only: [:show]
+
+  def show
+    @locale = params[:lang]
+
+    if I18n.available_locales.map(&:to_s).include?(@locale)
```
astronaut-wannabe: `I18n.available_locales` returns more locales than there are TCE yaml translations for.

```ruby
I18n.available_locales.map(&:to_s)
#=> ["en", "da", "de", "fr", "nb", "nl", "no", "pl", "pt-BR", "ru", "sv", "es", "si", "espanol"]
```

Maybe instead of using that method, I can hardcode in:

```ruby
TO_CHANGE_ANYTHING_YAMLS = ["espanol"].freeze
```

and then do
```ruby
 if TO_CHANGE_ANYTHING_YAMLS.include?(@locale)
...
...
```

app/views/layouts/to_change_everything.html.erb
astronaut-wannabe: @just1602 ah, ok 👍

app/controllers/to_change_everything_controller.rb
veganstraightedge: @astronaut-wannabe  Where is the list of langs coming from in this?

```ruby
I18n.available_locales.map(&:to_s)
```

astronaut-wannabe: @veganstraightedge it comes from `I18n.load_path`, which (at least on my local machine) includes some gem paths

You _can_ whitelist locales in the config like this:

```
# Whitelist locales available for the application
I18n.available_locales = [:en, :es, :espanol]
```

but that won't really solve the issue (which is that right now going to something like https://crimethinc-staging-pr-577.herokuapp.com/tce/en )

veganstraightedge: Gotcha. In that case, I think explicitly listing the TCE translations is a good approach.

`TO_CHANGE_ANYTHING_YAMLS = ["espanol"].freeze`

veganstraightedge:
Looks real good!

One small thing that's not right through is the lang nav on small screen / mobile.

Even when on `/tce/espanol`, it has English selected.

astronaut-wannabe:
> Even when on /tce/espanol, it has English selected.

ah! forgot about that. still trying to figure out a way to apply the css class and insert that `<span>` dynamically

app/assets/stylesheets/to_change_everything/_espanol.scss
```diff
@@ -0,0 +1,528 @@
+.espanol {
+    /*--- Highlight language in mobile language menu ---*/
+    #lang1 > li:nth-child(11) > a:nth-child(1) {
```
veganstraightedge: This seems fragile. All of these will be thrown off when we add new languages to the mix.

What if you change the `<span class="check>` to `<span class="check check-espanol>`?

Then this rule selector could be `.espanol .check-espanol`.

astronaut-wannabe: fixed

Issue #583, opened 2017-12-19
Re-layout "related articles" in Article footer


veganstraightedge:
- [x] Use a grid instead of a list (you'll need to new partial for this, instead of using `articles/list`)
- [x] Show `3` articles per category instead of `2` (on wide screens)
- [x] Hide the third article per category on smaller screens (maybe, below `$small-tablet`)

***

# Current

<img width="1086" alt="before" src="https://user-images.githubusercontent.com/4361/34142094-5c5769be-e429-11e7-836f-bf7544fd5b70.png">

# Expected

<img width="1052" alt="related articles" src="https://user-images.githubusercontent.com/4361/34142099-61f82e3a-e429-11e7-9df5-807a0ec63f85.png">

PR #594, opened 2017-12-25
Related articles redesign

just1602:
Please, check the CSS related to the `.related-article header` since it take me a while to have well place and align title and subtiltle, but I'm not sure if it's the best way to to it.


veganstraightedge:
screen shot 2017-12-28 at 8 21 43 pm

Good start, but it still needs work. Here are the issues I see:

- [ ] the whole `related articles` block is wider than the article content
- [ ] the `related articles` content is offset to the left of the `related articles` heading
- [ ] the `title` / `subtitle` is on top article header image. instead they should be underneath it. see, the "wireframe" in #583.
- [ ] the `title` / `subtitle` should be black text, on the white background of the page
- [ ] there should be more vertical space between each row of `related articles` categories

just1602:
The `.related-articles` follow the first paragraph width. Otherwise. I fix the things you ask.

veganstraightedge:
screen shot 2018-01-03 at 2 49 30 pm

veganstraightedge:
It looks like the main issue you're having is that each related item has a fixed height of 200px. Some title/subtitles are taller than that. Try allowing each article block (image, title, subtitle) to be as tall as it needs to be. It's ok if each item in the row is a different height.

veganstraightedge:
crimethinc how anti-fascists won the battles of berkeley 2017 in the bay and beyond a play-by-play analysis 2018-01-03 15-03-38

veganstraightedge:
Looks a lot better. And now it's *very* close to done.

I think this is the last issue.

The `Related Articles` section needs a little bit of white space around it. Right now when the window is narrow (but not mobile), the titles/subtitles run right up to the edge of the window. It should behave similar to the footer. Whatever that measurement is for the padding/margin on the footer, use that.

👍🏻

veganstraightedge:
Oops.

veganstraightedge:
screen shot 2018-01-07 at 12 10 08 pm
screen shot 2018-01-07 at 12 10 18 pm
screen shot 2018-01-07 at 12 12 41 pm

Issue #584, opened 2017-12-19
Use TCE assets on CDN


veganstraightedge:
I'm trying to clean house on `/public` as much as possible. To that end, and related to #575, Try to see if everything Just Works™ with TCE assets from the CDN.

- [ ] https://cloudfront.crimethinc.com/assets/tce/css
- [ ] https://cloudfront.crimethinc.com/assets/tce/js
- [ ] https://cloudfront.crimethinc.com/assets/tce/css/tcefont
- [ ] https://cloudfront.crimethinc.com/assets/tce/images

Also, because of historical CDN organizational reasons, can you:

- [ ] Use https://cloudfront.crimethinc.com/assets/tce instead of https://cloudfront.crimethinc.com/tce/

astronaut-wannabe:
#595 closes this **sort of**. 

only TCEs that use the erb template will no longer depend on the old assets. The remaining static html TCEs still rely on the original asset paths. 

**so we cannot delete the assets from /public just yet )**

veganstraightedge:
Copy that. I will not delete anything from `/public` in app or in `/tce` in S3 until all of the static files are migrated to the erb template and yaml translation. 👍🏻

PR #595, opened 2017-12-26
Use TCE assets on CDN

astronaut-wannabe:
I plan on "testing" this by doing a `diff -q old new` and making sure they are the same

- [x] `jquery.responsImg.min.js` diffed
- [x] `jquery.fancybox.css` diffed
  - This does differ, but when I ignore whitespace (`diff -qw old new`) there is no diff
- [x] `jquery.fancybox.pack.js` diffed
- [x] `video-play-icon-300.png` diffed
- [x] `<%=t('.pdf_image').html_safe%>` diffed
- [x] `circle-a-logo-hands-white-large@2x.png` diffed

closes [#584][0] (**sort of!** only TCEs that use the erb template will no longer depend on the old assets. The remaining static html TCEs still rely on the original asset paths. **so we cannot delete the assets from `/public` just yet** )

[0]:https://github.com/crimethinc/website/issues/584

### DIFF
--- a/app/views/layouts/to_change_everything.html.erb
+++ b/app/views/layouts/to_change_everything.html.erb
@@ -26,7 +26,7 @@
     }
   </script>
 
-  <script src="/tce/js/jquery.responsImg.min.js"></script>
+  <script src="https://cloudfront.crimethinc.com/assets/tce/js/jquery.responsImg.min.js"></script>
   <script>
     $(document).ready(function() {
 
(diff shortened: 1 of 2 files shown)


app/views/layouts/to_change_everything.html.erb
```diff
@@ -76,6 +76,7 @@
   <meta property="og:image" content="https://cloudfront.crimethinc.com/assets/tce/images/twitter-card-1486.jpg" />
   <meta property="og:type" content="article" />
 
+  <!-- TBD: fix path -->
   <link rel="stylesheet" href="https://cloudfront.crimethinc.com/tce/images/546676/CE039F8C7119876F1.css" />
```
astronaut-wannabe: @veganstraightedge I cannot find this file under the `https://cloudfront.crimethinc.com/assets/tce/css/tcefont` path

veganstraightedge: Copied it over.

`https://cloudfront.crimethinc.com/assets/tce/images/546676`…

astronaut-wannabe:
@veganstraightedge this should be good to merge. not sure why heroku is breaking though

PR #603, opened 2018-01-01
 [TCE] Add a first draft of a Persian version 

astronaut-wannabe:
We wanted to get a first draft of a farsiii version
live on the site ASAP due to some current events.

There probably still needs to be some work on polishing
the CSS and making the site fully RTL (for example, the menu
bar still reads LTR)

---

demo: https://crimethinc-staging-pr-603.herokuapp.com/tce/فارسی


astronaut-wannabe:
@veganstraightedge possible to get a staging deploy on this PR?

I think this is basically a serviceable rough-cut for production.

I will create a PR to track all of the polish that still needs to be done

app/views/layouts/to_change_everything.html.erb
```diff
@@ -127,7 +127,7 @@
               <li><a href="/tce">English</a></li>
               <li><a href="/tce/espanol">Español</a></li>
               <li><a href="http://www.crimethinc.com/blog/2016/09/21/to-change-everything-in-11-more-languages/#espanol">Español<br />&nbsp;&nbsp;&nbsp;(América Latina)</a></li>
-              <li><a href="http://www.crimethinc.com/blog/2016/09/21/to-change-everything-in-11-more-languages/#farsi">فارسی</a></li>
+              <li><a href="/tce/فارسی">فارسی</a></li>
```
astronaut-wannabe: this seems to render as valid html, I think it is correct and just looks like this due to the html being ltr and the text being rtl.... not sure though

veganstraightedge:
Images uploaded.

veganstraightedge:
There's a few tweaks that we should do for readability of text on top of images and the size of a couple, but this is pretty dang good enough to merge and launch. I'll make a separate issue for the tweaks.

PR #608, opened 2018-01-03
Fix the facebook share url

just1602:


just1602:
So, for a very weird reason, when we use `&amp;` instead of the `&` character, the link doesn't work anymore. In [the documentation](https://developers.facebook.com/docs/sharing/reference/feed-dialog) facebook use the `&amp;` name.

When we use this url `https://www.facebook.com/dialog/feed?app_id=966242223397117&amp;display=popup&amp;caption=How%20Anti-Fascists%20Won%20the%20Battles%20of%20Berkeley&amp;link=https://crimethinc.com/2018/01/03/how-anti-fascists-won-the-battles-of-berkeley-2017-in-the-bay-and-beyond-a-play-by-play-analysis&amp;redirect_uri=http://www.facebook.com/&amp;display=popup` we got this page : 

screen shot 2018-01-03 at 6 07 07 pm

When we use this url `https://www.facebook.com/dialog/feed?app_id=966242223397117&display=popup&caption=How%20Anti-Fascists%20Won%20the%20Battles%20of%20Berkeley&link=https://crimethinc.com/2018/01/03/how-anti-fascists-won-the-battles-of-berkeley-2017-in-the-bay-and-beyond-a-play-by-play-analysis&redirect_uri=http://www.facebook.com/&display=popup` we got this page : 

screen shot 2018-01-03 at 6 10 42 pm

The second one content the crimethinc page preview (or object as facebook call it) and let the user post it. The first one didn't and if you try to only publish text, it give you an error.

@veganstraightedge @astronaut-wannabe have you already seen encoding (i guess)( problems like that?

If we use `&` instead of `&amp;` what are the problem or security flaw we can face?

veganstraightedge:
@just1602 Sorry. This is my fault. It's just an old habit of encoding all `&`s in HTML as `&amp;`, even ones in URLs. Historically, it's worked for me. But I didn't notice it caused this to fail.

Good catch, @just1602! ⭐️

PR #612, opened 2018-01-04
 [TCE] Update header font for farsi

astronaut-wannabe:
- Use the same font for the headers that the PDF version uses
  - [Noto Kufi Arabic][0]
- Update the headers to have `font-weight: 900`
- Update `line-height` of headers
- tweak some paddings/margins

[0]:https://www.google.com/get/noto/#kufi-arab


astronaut-wannabe:
@veganstraightedge this PR fixes some of the hard-to-read headers for the farsi and is good to ship after any tweaks you see needed.

I will be moving on to the `<p>` body text next.

I load the font from google's server. so maybe you would want to upload the file to the CDN instead: https://fonts.googleapis.com/earlyaccess/notokufiarabic.css

PR #643, opened 2018-01-29
Change the share fb btn to use the sharer link

just1602:
I don't know why, but I start to receive error again when I want to share an article on facebook using the facebook share button. I discover that there is a facebook service call sharer that we can call using `https://facebook.com/sharer?u=https://test.com` or `https://facebook.com/sharer/sharer.php?u=https://test.com`. I'll use this service since we didn't need to have an app_id or anything else like that.


app/helpers/articles_helper.rb
```diff
@@ -13,7 +13,7 @@ def social_link_for article, site
     when :twitter
       "https://twitter.com/intent/tweet?text=#{url_encode article.title}&amp;url=#{article_url}&amp;via=crimethinc"
     when :facebook
-      "https://www.facebook.com/dialog/feed?app_id=966242223397117&display=popup&caption=#{url_encode article.title}&link=#{article_url}&redirect_uri=#{article_url}"
+      "https://www.facebook.com/sharer?u=#{url_encode article_url}"
```
astronaut-wannabe: is it possible to do something like `url_encode article_short_url || article_url`

just1602: Yeah, while walking back home I was wondering if it could be possible to do something like that.

astronaut-wannabe: @just1602 ahh, I see. then yea, this makes more sense

Issue #598, opened 2017-12-29
Redirect from Heroku subdomain


veganstraightedge:
# Steps

Go to: https://crimethinc.herokuapp.com

# Actual

Left at https://crimethinc.herokuapp.com

# Expected

Redirect to: https://crimethinc.com

PR #644, opened 2018-01-30
[#598] Redirect from heroku subdomain

just1602:
Now redirect from crimethinc.herokuapp.com to crimethinc.com

It's quite simple, but maybe extract the regex comparaison in a private method `is_domain_matching request, domain` to avoid code duplication, but otherwise it should work.


just1602:
It works so well, that it redirect you to `https://crimethinc.com` when you try to hit `https://crimethinc-staging-pr-644.herokuapp.com/`.

Issue #625, opened 2018-01-10
Add GitHub icon to social media icons in footer


veganstraightedge:

astronaut-wannabe:
https://github.com/edent/SuperTinyIcons

PR #645, opened 2018-02-01
Add GitHub icon to social media icons in footer

Butch3rPing:
In reference to #625. 

I looked at the repo that @astronaut-wannabe linked to but only found a rectangle icon which sorta breaks the flow. A circle would probably be better but I'm not sure where y'all got the other icons

git

😃


astronaut-wannabe:
@Butch3rPing good catch!

maybe I got the others from here? https://github.com/gilbarbara/logos

There is a circle github icon in there you can download in that repo

Butch3rPing:
Perfect.
git

veganstraightedge:
@Butch3rPing Great job! Thanks for diving in. 🌟⭐️🌟⭐️

Issue #602, opened 2017-12-31
Redirects to external URLs aren't saving the URL


veganstraightedge:
Create a new Redirect with any `source_path` and a `target_path` of `http://example.com/foo/bar`.

```ruby
redirect = Redirect.create(source_path: "foo", target_path: "http://example.com/foo
```

# Actual

```ruby
redirect.target_path

# => /example.com/foo
```

# Expected

```ruby
redirect.target_path

# => http://example.com/foo
```

PR #647, opened 2018-02-03
[#602] fix Redirect url parsing and saving bug

just1602:
This fixed the #602 bug.

Maybe just check the `groomed_path_or_url` method diff. I had weird bug, so when I tried to have something like `url_pieces << "#{url.scheme}://#{url.host}" unless url.host =~ %r{crimethinc.com|cwc.im}` the `://` section was append to the parts even if the regex catch our domains.

If you think it's just me and there is something to do to make the method cleaner and avoid the `unless / else` statements tell me, I'll investigate more on that. Otherwise, it works now.


just1602:
I found the bug, it was because the regex wasn't catching empty domain and also because URI doesn't return an empty string, but `nil` when there is no host.

veganstraightedge:
@astronaut-wannabe Can you review this one please?

app/models/redirect.rb
```diff
@@ -50,20 +50,16 @@ def paths
...
 
   def groomed_path_or_url url
-    url_pieces = []
-    url_pieces << url.host unless url.host =~ %r{crimethinc.com|cwc.im}
-    url_pieces << url.path
-    url_pieces << "?" + url.query    if url.query.present?
-    url_pieces << "#" + url.fragment if url.fragment.present?
-    url_pieces.join
+      url_pieces = []
+      url_pieces << "#{url.scheme}://#{url.host}" unless url.host =~ %r{crimethinc.com|cwc.im} or not url.host.present?
```
astronaut-wannabe: the `unless` + `or not` combo is confusing me.

I think using `not something.present?` is the same as `something.blank?`

doing that substitution results in:
```ruby
url_pieces << "#{url.scheme}://#{url.host}" unless url.host =~ %r{crimethinc.com|cwc.im} or url.host.blank? 
```
now when I read this, I assume it is saying 

> add this string to `url_pieces` unless either one of these conditions is true:
>  1. it is a crimethinc url 
>  2. url_host is `nil`

if that is what it is supposed to do, I'd prefer:
```ruby
url_pieces << "#{url.scheme}://#{url.host}" unless crimethinc_url?

#...

def crimethinc_url?
  url.host.blank? || url.host =~ %r{crimethinc.com|cwc.im} 
end
```

just1602: It's stupid since I didn't know about the `blank?` method, next time I'll know. For the `crimethinc_urls?` method, it's really cleaner that way !

app/models/redirect.rb
```diff
@@ -50,20 +50,20 @@ def paths
   end
 
   def build_url_for path
-    URI.parse("http://" + strip_protocol_from_path(path))
```
astronaut-wannabe: I am not exactly clear why we were doing `strip_protocol_from_path` in the first place. It seems like it just removed any `http` or `https` and then just immediately replaced it with an `http`...

just1602: The only reason I saw was because the regex in `add_leading_slash` method was only handling `http`. Now it handle both `http` and `https`. 

So I refactor this because : 

1. The method seems to be pretty useless now
2. It remove to possibility to redirect to a secure link

Maybe  @veganstraightedge  can explain us if there any other reasons to strip the protocol and always keep it to `http`.

veganstraightedge: Feel free to tear this all down and re-build it if yall wanna. My implementation was very quick and very dirty. I'm sure it can be done in a better way.

astronaut-wannabe: @veganstraightedge @just1602 in that case, I say just ship this. there is no real need to refactor right now (the code is not coupled to anything else, it is not blocking other features, and there haven't rally been any bugs besides this one)

I would prefer to just leave it be and let the test coverage continue to build up around any other edge cases that may come up. That way, if we *do* refactor we can do it with confidence.

Also, right now 2 people are pretty familiar with this code (@just1602 + @veganstraightedge), after a refactor it would be back down to 1

Issue #110, opened 2016-12-19
Animate nav background color


veganstraightedge:
Real slow fade between a handful of colors.

Not sure what those colors are yet, how many there will be or how fast it should fade.

So, let's put all those things in SCSS variables/functions, so we can easily iterate it.

*** 

See: https://sixcolors.com for an example. We'd use different colors and way slower. Probably.

astronaut-wannabe:
closed via #654

PR #654, opened 2018-02-10
Animate nav background color #110

TheConObs:
Adds a color fader mixin and the associated style updates, including a small tweak to the header nav < 800px now that the logo bar won't be black.


app/assets/stylesheets/components/_header.scss
```diff
@@ -1,10 +1,19 @@
...
   align-items: center;
-  background-color: $color-true-black;
+  background-color: #76c261;
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
   padding: $border-size-medium $border-size-medium;
+  // #110 ~ Play with the duration to get a timing that feels right
+  //        depending on the number of colors set in $header-color-list
+  -webkit-animation: colorfader linear infinite;
```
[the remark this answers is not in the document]
TheConObs: Does rails-autoprefixer do that for the keyframe declarations, too?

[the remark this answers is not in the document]
veganstraightedge: Not sure! If you remove the prefix and then refresh. You should be able to inspect the generated CSS to see if it does or not.

app/assets/stylesheets/components/_navigation.scss
```diff
@@ -2,7 +2,10 @@
 .primary-navigation {
   display: flex;
   flex-wrap: wrap;
-  background-color: $color-true-black;
```
astronaut-wannabe: This is a design question for @veganstraightedge 

do we want to add a change like this:
```diff
diff --git a/app/assets/stylesheets/components/_navigation.scss b/app/assets/stylesheets/components/_navigation.scss
index e3a0c2be..689318c3 100644
--- a/app/assets/stylesheets/components/_navigation.scss
+++ b/app/assets/stylesheets/components/_navigation.scss
@@ -2,9 +2,17 @@
 .primary-navigation {
   display: flex;
   flex-wrap: wrap;
-  background-color: $color-true-black;
+  animation: colorfader linear infinite;
+  animation-duration: 270s;
+  animation-iteration-count: infinite;
+
+  @media screen and (max-width: $small-tablet) {
+    background-color: $color-true-black;
+  }
 }
 
+@include color-fader("colorfader", $header-color-list);
+
 .nav-link {
   font-size: 1.7rem;
   letter-spacing: .05em;
```

so that when the screen is a small display and the nav-header is on 2 lines the top and bottom row both cycle?

e.g.

only top row             |  both rows
:-------------------------:|:-------------------------:
screen shot 2018-02-15 at 13 48 44 | screen shot 2018-02-15 at 13 42 29

veganstraightedge: Good catch @astronaut-wannabe. :shipit:

TheConObs: Nice! I didn't even think about that.

TheConObs: Is that tweak something you're adding to the mix, @astronaut-wannabe? Just want to make sure I'm not dropping the ball.

astronaut-wannabe: @TheConObs sorry, I totally missed your question!

I was suggesting that you add that change to this PR

TheConObs: Ah OK. Upon some initial low-level tinkering this morning, it will take some tweaking but I'll have this up this weekend.

app/views/shared/_header.html.erb
```diff
@@ -1,3 +1,2 @@
 <%= render "shared/header/masthead" %>
-<%= render "shared/header/nav" %>
```
astronaut-wannabe: what does this do?

TheConObs: So currently in production /nav is the 2nd row nav menu that appears on smaller devices. In order for the fader to work for the menu, I had to remove the call to that 2nd nav and adjust the main centered nav to stay visible but move beneath the logo.

Adding the fader to that /nav menu didn't work as the two rows being separate containers wouldn't be in sync. Plus this way adds the benefit of being DRY-compliant since there will no longer be a duplicated header menu in the DOM just waiting for a media query to kick in.

veganstraightedge:
What's the status of this PR? I've been MIA awhile. So, I don't have a lot of context on it. Sorry.

astronaut-wannabe:
@veganstraightedge , I think it is good to go (right, @TheConObs )?

the only thing that might be left is [addressing this comment][0] about finalizing the color choice

[0]:https://github.com/crimethinc/website/pull/654/files#diff-28ea3a81b70aadbbba4580c4a2cade60R11

veganstraightedge:
@astronaut-wannabe Thanks.

@TheConObs If that's the only thing left, will you make all of the colors black for now please? That way we can ship this and then just swap out the colors when they're ready.

TheConObs:
Indeed. It's good to go, aside from as you mentioned, color choices. We can obviously either set them now if anyone knows what they should be, or do it in later.

TheConObs:
Ok, yeah. So... disregard my last comment.

PR #665, opened 2018-02-19
Remove pinning

astronaut-wannabe:
pinning was added in https://github.com/crimethinc/website/pull/524 to make sure development matched prod.

prod was updated in c2e26562090026641e601481b50dc44a53812c2c, so pinning is no longer needed


astronaut-wannabe:
@veganstraightedge @eliduke per the slack discussion

PR #690, opened 2018-03-29
Form tags

TheConObs:
Updates form_for / form_tag to form_with.


TheConObs:
I don't have permission to view the issues that are apparently coming up under snyk. What's happening?

veganstraightedge:
@TheConObs Don't worry about them. Unfortunately they are upstream issues that we don't control. :/

GeoffFoley:
Ok, the damned color-list variable. I'll fix as soon as I'm home.

guitsaru:
Are the `PGDATA` files related to this PR?  There's a lot of newly created binary files.

guitsaru:
@veganstraightedge any idea why the heroku status check hasn't reported yet?

astronaut-wannabe:
@guitsaru I noticed that seems to happen whenever the PR is across forks (as opposed to across branches)

just1602:
Yeah,if I remember well, when you're not in the team and cannot branch or push directly to the project, @veganstraightedge has to lunch the deployment manually.

guitsaru:
Ah, yep, that makes sense.

GeoffFoley:
I found that it was set to true globally, although I don’t remember the file name at the moment, so I removed it from forms that I could but IIRC some forms were still set up in a way that they stopped working if I took it out.

I’m not sure why that was happening as I’m new to Rails.

veganstraightedge:
Thanks for diving into this one @TheConObs! And sorry the process took so long.

PR #760, opened 2018-06-13
Donations with Stripe credit card processing!

eliduke:
This PR is getting pretty big, but here are the two main pages that are being added.

The Main Support page where you can give a one-item donation, sign up for a monthly subscription, and also request a link to make changes to an existing subscription:

screen shot 2018-09-18 at 12 31 24 pm

Email with link to Support Edit:

screen shot 2018-09-18 at 1 17 17 pm

The Support Edit page where you can update or cancel a subscription:

screen shot 2018-09-18 at 12 30 55 pm


app/views/donations/new.html.erb
```diff
@@ -14,15 +20,42 @@
...
-          data-key="<%= Rails.configuration.stripe[:publishable_key] %>"
-          data-description="A month's subscription"
-          data-amount="5000"
-          data-zip-code="true"
-          data-locale="auto"></script>
-<% end %>
+  <button id="js-stripe-button">DONATE</button>
+
+  <script>
+    var handler = StripeCheckout.configure({
```
eliduke: If anyone has any ideas as to how to get this out of the view, I'm _all ears_. The problem is that the:
```
<script src="https://checkout.stripe.com/checkout.js"></script>
```
Near the top of this file has already loaded and allows for calling `StripeCheckout.configure`. I'm not sure how to extract everything in the `<script>` block out into the `donations.js` while still allowing for access to `StripeCheckout`.

app/views/donations/new.html.erb
```diff
@@ -1,4 +1,10 @@
-<%= form_tag [:donations] do %>
+<%= form_tag [:donations], id: "js-stripe-form" do %>
```
just1602: I don't know for form that are for external library or other thing like that, but we try to use `form_with` for the form. If it's specific to Stripe or there is something that you need to do that can be do with `form_with` it's cool. Otherwise, consistency would be nice. :)

eliduke: Totally cool! I've been in Rails 4.2 land for a while now with work and just wasn't thinking about it. I'm sure that `form_with` should be just fine. 👍

just1602:
On the support page, the link in the about crimethinc div in the footer are also green. :P

just1602:
@veganstraightedge I'd put `font-weight: 400` on the next button like the mothly / one-time label and I'd also set the mouse cursor to `pointer` when you hover the next button.

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        customer: customer.id,
+        plan:     STRIPE_MONTHLY_PLAN_ID,
+        quantity: amount
+      )
+    else
+      Stripe::Charge.create(
+        customer:    customer.id,
+        amount:      amount.to_i * 100,
+        description: t('views.donations.descriptionl_one_time'),
+        currency:    'usd'
```
astronaut-wannabe: should this come from the front-end in case someone is more comfortable with a different local currency

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def thanks
+    @html_id = 'page'
+    @body_id = 'support'
+    @title   = I18n.t('views.donations.thanks.heading')
+  end
+
+  def create
+    amount = params[:amount].tr('$', '')
+
+    customer = Stripe::Customer.create(
```
astronaut-wannabe: I am having a hard time figuring out the answer to this from the stripe docs, so maybe i will just experiment, but:

Is this a `find_or_create` type operation? if someone donates a 1-time donation, we add them as a customer. If they come back to do another donation, will we get an error, or will everything work fine?

app/views/donations/new.html.erb
```diff
@@ -0,0 +1,88 @@
...
+        <div class='column column-half'>
+          <%= radio_button_tag :monthly, false %>
+          <%= label_tag :monthly_false, t('views.donations.support.radio_label_one_time') %>
+        </div>
+      </div>
+
+      <%= range_field_tag 'amount_slider', 50, in: 1..1000, step: 1 %>
+
+      <%= label_tag :amount, t('views.donations.support.amount_label') %>
+      <%= text_field_tag :amount, '$50', tabindex: '-1' %>
```
astronaut-wannabe: this feedback may be overkill for the initial implementation, so feel free to defer for now:

this is making me nervous. Does strip have a frontend UI library?

If not, maybe make it a `<input type="number" step="1" min="0" max="10000">` and use javascript (Jquery or vanilla) to make sure the backend just gets an integer value in cents and a currency code

app/controllers/donations_controller.rb
eliduke: Good question! I think for now we can just roll with `usd` and maybe have other currencies be a donations v2 feature. Anyone else have an opinion about that?

app/controllers/donations_controller.rb
eliduke: This came up while @veganstraightedge and I were working through some of this and we decided to just create new customers each time. If someone wants to give a one time donation and then create a monthly one, we'll just create a new customer. If that gets out of hand and we get a bunch of duplicate customers then maybe we can adjust that in the future. Personally, not too worried about it.

app/controllers/donations_controller.rb
veganstraightedge: I'm fine with being USD only for now. We'll let people tell us that they want local currency support to figure out if we need to build it.

app/views/donations/new.html.erb
eliduke: I have a solution to this that I am just wrapping up right now.

It cleans up some of the js where we were adding and removing $ and sets the field to a number field. It did require that I move the "amount" to being left aligned. It was much easier that way and I think it's a cleaner implementation. Pushing it up in a sec.

eliduke:
@veganstraightedge I looked into what it takes to "activate" a stripe account and it's pretty in-depth:

strip activation

app/controllers/donations_controller.rb
```diff
@@ -41,4 +43,22 @@ def create
   else
     redirect_to [:thanks]
   end
+
+  def stripe_webhooks
+    event_json = JSON.parse(request.body.read)
+
+    if event_json["type"] == "invoice.payment_succeeded"
+      customer_id = event_json["data"]["object"]["customer"]
+      customer  = Stripe::Customer.retrieve("cus_DCbYHIi93Ldej7")
```
just1602: We shouldn't retrieve from the `customer_id` instead of an hardcoded id?

eliduke:
@just1602 I saw your comment about the hard-coded id and, yes, you are totally right. That was a goof and I got that fixed. 👍

app/controllers/donations_controller.rb
```diff
@@ -64,9 +64,9 @@ def create_session
 
         SubscriptionMailer.with(subscription: subscription, email: email).edit.deliver_later
 
-        flash[:notice] = "We sent you an email with a link to do the thing you need to do."
+        flash[:notice] = 'We sent you an email with a link to do the thing you need to do.'
       else
-        flash[:error] = "We found multiple subscriptions with that email address."
+        flash[:error] = 'We found multiple subscriptions with that email address.'
```
veganstraightedge: Can you move both of these strings into `en.yml` please?

app/controllers/donations_controller.rb
```diff
@@ -95,12 +95,12 @@ def update_subscription
...
       if subscription&.save
-        flash[:notice] = "Your subscription amount has been updated!"
+        flash[:notice] = 'Your subscription amount has been updated!'
       else
-        flash[:error] = "There was a problem canceling your subscription."
+        flash[:error] = 'There was a problem canceling your subscription.'
       end
     else
-      flash[:error] = "You have to select an amount to update."
+      flash[:error] = 'You have to select an amount to update.'
```
veganstraightedge: To `en.yml` also

app/controllers/donations_controller.rb
```diff
@@ -110,9 +110,9 @@ def cancel
     subscription = Stripe::Subscription.retrieve(params[:subscription_id])
 
     if subscription&.delete
-      flash[:notice] = "Your subscription has been canceled."
+      flash[:notice] = 'Your subscription has been canceled.'
     else
-      flash[:error] = "There was a problem canceling your subscription."
+      flash[:error] = 'There was a problem canceling your subscription.'
```
veganstraightedge: `en.yml`

spec/mailers/previews/subscription_mailer_preview.rb
```diff
@@ -3,7 +3,7 @@ class SubscriptionMailerPreview < ActionMailer::Preview
   def edit
     SubscriptionMailer.with(
       subscription: SubscriptionSession.first,
-      email: "test@gmail.com"
+      email: 'test@gmail.com'
```
veganstraightedge: If this is just for testing purposes, let's use `example.com`

app/controllers/donations_controller.rb
eliduke: Totally. I left all text out of `en.yml` for now because everything still feels a little TBD. My plan was to wait until things are exactly where/how we want them and _then_ add everything there.

spec/mailers/previews/subscription_mailer_preview.rb
eliduke: 👍

veganstraightedge:
You could do that. Or you could add a file specific exclude for that type of warning in `.rubocop.yml`

eliduke:
@veganstraightedge Hey hey heeeey! 🕺

Looks like we've still got two issues on code climate regarding that long helper method that I made an exception for in rubocop, and also the complexity of the `create_session` method:

```
def create_session
  email = params[:email]
  customers = Stripe::Customer.list(email: email).data

  if customers.present?
    customers.each do |customer|
      customers.delete(customer) if customer.subscriptions.data.empty?
    end

    if customers.empty?
      flash[:error] = "We can't find a subscription with that email address."
    elsif customers.one?
      subscription = SubscriptionSession.create!(
        stripe_customer_id: customers.first.id,
        token: SecureRandom.hex,
        expires_at: 1.hour.from_now
      )

      SubscriptionMailer.with(subscription: subscription, email: email).edit.deliver_later

      flash[:notice] = 'We sent you an email with a link to do the thing you need to do.'
    else
      flash[:error] = 'We found multiple subscriptions with that email address. Please send us an email.'
    end
  else
    flash[:error] = "We can't find a subscription with that email address."
  end

  redirect_to [:support]
end
```

I agree that it feels too complex, too many layers to that onion, but I don't really know how to clean it up. There are just so many scenarios that need checking and I can't think of how to make it easier to read. Maybe a `case` statement somewhere? /shrug

Another questionable spot:

```
def find_or_create_customer(email:, source:)
  return if email.blank? || source.blank?

  customers = Stripe::Customer.list(email: email).data

  if customers.present?
    customers.first
  else
    create_customer(email, source)
  end
end
```

Because `store.crimethinc.com` is creating new a Stripe 'customer' for _every_ order (regardless of repeat email), we have lots customers with the same email address. I wasn't sure how best to address that when doing a `find_or_create` sorta thing. I opted to just grab the first customer because that is the most recent one. Thoughts on that?

Also, a couple other follow-up questions:

1. Do you still want me to rename: `DonationsController => SupportController`?
2. Should we limit people to just ONE cwc subscription per email / customer?
3. How do you feel about the general approach to the support edit screen:

Without an 'amount' selected:
Update without amount selected

With an 'amount' selected:
Update with amount selected

I did a super quick styling of things, totally open to suggestions. The main issue for me was creating as few forms as possible. It seemed a little too much to create _another view_ for a single drop-down update form. Shoving the 'update' form into the table seemed like a happy compromise to keep all subscription "actions" in one place.

The credit card update is not functional yet. I was thinking about _maybe_ holding off on that for now, button up what we've got elsewise. This is starting to feel like a pretty big PR and it would be nice to carve a little bit out so that I (we) can focus on refactoring what needs refactored and get this up to production! The credit card update functionality could always be an immediate follow-up PR.

I still need to move all the inline text to `en.yml`, but I want to wait until we've refactored what we want and have a very clear picture of what text we actually need. Oh, and @just1602, I tried converting the `form_tag` to `form_with`, but the Stripe checkout stuff was being weird and not populating the hidden fields like it had been:

```
<%= hidden_field_tag 'js-stripe-email' %>
<%= hidden_field_tag 'js-stripe-token' %>
```

I saved a copy of my attempts with `form_with`, so it could give it another shot later on, but it was starting to frustrate me. 🤦‍♂️ Maybe we can punt that refactor to the follow-up PR as well?

Anyhoooooo, lemme know how y'all want to proceed.

app/helpers/support_helper.rb
```diff
@@ -1,4 +1,15 @@
 module SupportHelper
+
+  def time_until_expiration(subscription_session)
+    minutes = (subscription_session.expires_at.to_i - Time.current.to_i) / 60
+
+    if minutes == 0
```
veganstraightedge: `if minutes.zero?`

app/mailers/subscription_mailer.rb
```diff
@@ -2,7 +2,8 @@ class SubscriptionMailer < ApplicationMailer
   def edit
     @email = params[:email]
     @subscription = params[:subscription]
+    @host = params[:host]
 
-    mail to: @email, subject: 'Request to update Crimethinc. Support'
+    mail to: @email, subject: 'Request to update your CrimethInc Support'
```
veganstraightedge: `CrimethInc.`

app/views/subscription_mailer/edit.html.erb
```diff
@@ -3,8 +3,8 @@
 <p>Click this link to edit your support:</p>
 
 <p>
-  <%= link_to support_edit_url(host: 'crimethinc.com', token: @subscription.token),
-              support_edit_url(host: "crimethinc.com", token: @subscription.token) %>
+  <%= link_to support_edit_url(protocol: 'https', host: @host, token: @subscription.token),
+              support_edit_url(protocol: 'https', host: @host, token: @subscription.token) %>
```
veganstraightedge: Maybe you can set `protocol` as a `default` on the route itself.

https://guides.rubyonrails.org/routing.html#defining-defaults

app/helpers/support_helper.rb
eliduke: Yeah, Rubocop barked at me about that. Just haven't push it up yet.

app/views/subscription_mailer/edit.html.erb
eliduke: Oooooo 🤔

eliduke:
I refactored the support edit page:

sub edit 1

sub edit 2

eliduke:
@veganstraightedge I'm getting pretty close here. I still need to flesh out the emails a tiny bit, move all the new text into `en.yml`, and figure out how to simplify the newly complicated `create` method.

app/controllers/support_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
     if params[:monthly] == 'true'
       if customer_with_subscription(email)
         flash[:error] = "We already have a monthly subscriber with that email address. If you'd still like to give more, try a one-time donation. Thanks!"
-        redirect_to [:support] and return
+        redirect_to([:support]) && return
```
just1602: redirect_to doesn't return by itself ?

config/environments/production.rb
```diff
@@ -40,7 +40,7 @@
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  config.assets.js_compressor = Uglifier.new(harmony: true)
```
eliduke: https://github.com/lautis/uglifier/issues/127#issuecomment-352224986

eliduke:
@veganstraightedge If you've got some time in the next couple days, we could take a look at the newly-complex `create` method. 🙄

Issue #480, opened 2017-09-13
Create text only site


veganstraightedge:
In the style of these sites:

- http://thin.npr.org
- http://lite.cnn.io

Let's make a text only version of these pages:
- home page
- article page
- tag page
- category page
- paginated versions of each of those 

If someone other than me takes on this issue, let's talk strategy before diving in.

TheConObs:
Have you taken this one on yet? Looks like it could be a fun one to pull off.

veganstraightedge:
@jennifermf is working on it.

socialecologist:
Initial considerations/questions:
- subdomain or separate folder?
- serve dynamically or generate static files, with a job?
- support/considerations for translations?
- re: tag page. Does this page currently exist? I could find /categories for that page?

jmfcodes:
@socialecologist those are important considerations. i'm not sure how to address the first and third.

for the second... since text files are so small. i wondered if the text-only version could be generated automatically when the non-text version is created.

currently, each article on the main page has a `<style="background-image: url(https://cloudfront.crimethinc.com/assets/articles/2018/07/11/header.jpg)">` and i think for maximum accessibility, it would be good to include image alt-text. however, i don't think alt-text exists here on the main page.

socialecologist:
Also need to figure out how to handle the routes. Right now the PR has `/txt` and `/txt/:article_id` for two of the pages.

socialecologist:
re: background images. You'll see I'm just using a new layout for the text only pages, so those aren't showing up. Some of the articles however have html content format with `img` tags in the content, so those ones are showing up right now.

socialecologist:
Yet another thing to consider is how to link to this from the main navigable piece of the website. Could add something to the main footer.

veganstraightedge:
Hi @socialecologist. Thanks for getting into this.

Here's my opinion on what strategy would be best.

> subdomain or separate folder?

Subdomain. 

> serve dynamically or generate static files, with a job?

For now, dynamically like the image-full site. Generate static files is outside the scope of this issue. And if we did do that, it would also benefit doing it for the image-full site. It's not a slam dunk though. Re-generation / replacing old versions if they get updated, where to store them, where to serve them from, deployment process… all of these are trade-offs that come with moving to static site germination. So, for now, let's punt on this idea.

> support/considerations for translations

For now, we don't have translations. But they're _hopefully_ coming in the next couple months.

[...]

PR #801, opened 2018-07-20
Text only pages start, home page

socialecologist:
This is work towards #480.

Throwing this up, not finished yet, but wanted to discuss.


veganstraightedge:
This branch/PR is pretty stale and I think we're going a different direction for doing a text only / lite site. So, I'm closing out this PR. Thanks for the start on this project though!

Issue #773, opened 2018-06-23
Add image uploads functionality to Article from in /admin CMS


veganstraightedge:
Right now, a person has to manually:
- resize (dimensions and resolution) images
- strip exif metadata images
- rename  images
- create folders on S3 to match the DD in the MM in the YYYY
- upload each of the article images
- upload the header image
- place the images in the article using our `[[ MEDIA URL ]]` syntax
- copy the `header.jpg` image URL from S3 and paste it into the header image field

What would be a bajillion times better is something like:

- someone in the content side find and/or edit images for content/color/crop
 - but not resolution, dimension, exif, compression
- then they’d drag them into the article form in the browser
- the images would get ajax uploaded in the background to a folder on S3 named something (more on that below)
- then the URLs and a small thumbnail of the images would be displayed in the form for copy/pasta into the article content (better UX could be added later)
- the backend would handle the resolution, dimension, exif, compression concerns

What to name that folder on S3?

- I’d LOVE it to be the DD of the article, but that changes sometimes while an article is in draft
- It could be the article slug (title/subtitle), but that changes too in the draft process
- Same for the short URL (though this is probably the safest)

And it only has to be unique with the MM on S3

Rails 5.2 added ActiveStorage. I think that’d be perfect to use for this.

http://guides.rubyonrails.org/active_storage_overview.html

https://www.schneems.com/2018/05/11/rails-52-active-storage-previews-poppler-and-solving-licensing-pitfalls/

https://weblog.rubyonrails.org/2018/4/9/Rails-5-2-0-final/

https://weblog.rubyonrails.org/2017/11/27/Rails-5-2-Active-Storage-Redis-Cache-Store-HTTP2-Early-Hints-Credentials/ (edited)

That alone would reduce the friction for the publishing side a great deal

veganstraightedge:
Many moons later, we’re now thinking about this functionality differently. 

The current workflow used by content folx, as described above, is still how things are done. 

We still want an uploader, for content folx’ self service. 

But, instead of having the uploader component on the admin article form, it’ll be on its own page. 

Probably at: `/admin/upload`

And a separate page/sections of pages for using the uploaded media in a media library. 

`/media`
`/library`
`/assets`

I’m personally leaning toward `/media`. At least partially because we gem that’s used to embed things in articles is called: `markdown_media`. 

https://github.com/veganstraightedge/markdown_media

Uploader page should support:

- All of the formats that active storage supports 
- drag and drop
- multiple files dropped
[...]

PR #812, opened 2018-08-01
[WIP] Add article header image upload via ActiveStorage

astronaut-wannabe:
relates to: https://github.com/crimethinc/website/issues/773

First, sorry this has taken so long

## The Goal of this change
I just want to get the wiring between ActiveStorage and AWS connected and working in prod before building out the rest of the features (the in-article image uploading).

To do this I added a header image upload form to the admin page which we can use to try out ActiveStorage in production (the original method of setting the header image is still the default)

## Some comments after learning ActiveStorage
I have learned some things about ActiveStorage that I realize will conflict with some of the specs in #773. I want to point them out now to find out if they are blockers we need to find work-arounds for or not

### custom asset URLs
from #773 
> What to name that folder on S3?
>
>    I’d LOVE it to be the DD of the article, but that changes sometimes while an article is in draft
>    It could be the article slug (title/subtitle), but that changes too in the draft process
>    Same for the short URL (though this is probably the safest)
>
> And it only has to be unique with the MM on S3

Rails ActiveStorage actually doesn't let you set the url. The reason for this is because [from the docs][0] is:

> Generate a permanent URL for the blob that points to the application. Upon access, a redirect to the actual service endpoint is returned. ***This indirection decouples the public URL from the actual one, and allows, for example, mirroring attachments in different services for high-availability***. The redirection has an HTTP expiration of 5 min.

So, activestorage doesn't allow for custom URLs as they are all managed by Rails and tracked in the database
### article fetching
ActiveStorage requires 3 database queries to resolve a URL. This seems like it could be an issue when I add the in-article image features (both in the rendered article and in the admin UI)

It is something to keep in mind, but there seems to be ways to mitigate it (https://evilmartians.com/chronicles/rails-5-2-active-storage-and-beyond)

## Explanation of code changes
I will go through and annotate some of the code change to explain what is going on 

[0]:https://edgeguides.rubyonrails.org/active_storage_overview.html#linking-to-files


app/views/articles/_image.html.erb
```diff
@@ -1,7 +1,19 @@
-<% if article.image.present? %>
+<% if article.header_image.attached? %>
```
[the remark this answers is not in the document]
jmfcodes: responsive design! i like it.

astronaut-wannabe:
> Regarding ActiveStorage not creating variants until the first time they're accessed: does this affect load times for users on slower/less reliable connections in any measurable (read: frustrating for users) way? I imagined it would be negligible/unnoticed, especially considering it's a one-time thing (per article/variant)-- so I was mostly just curious.

@jmfcodes i believe the answer to this is "yes"

though, I am just now reviving this branch so i will need to see if there have been any updates to activestorage and also we could always force-create the variants before the article is published somehow

config/storage.yml
```diff
@@ -0,0 +1,14 @@
...
+test:
+  service: Disk
+  root: <%= Rails.root.join('tmp/storage') %>
+
+amazon:
+  service: S3
+  access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
+  secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
+  region: 'us-west-2'
+  bucket: 'cwc-website'
```
veganstraightedge:
```suggestion
  bucket: <%= ENV.fetch('AWS_BUCKET') { 'cwc-website' } %>
```

Can you change this to read from an ENV with a fallback default?

app/views/articles/_image_media.html.erb
```diff
@@ -1,7 +1,19 @@
...
+       <%= url_for(image_variant_by_width(article, 640)) %> 640w,
+       <%= url_for(image_variant_by_width(article, 320)) %>  320w"
+       >
+</div>
+<% elsif article.image.present? && media_mode? %>
   <div class="article-image">
     <%= link_to_if linked,
-                   image_tag(article.image, class: "u-photo", alt: article.image_description),
-                   article.path %>
+        image_tag(url_for(article.image), class: "u-photo", alt: article.image_description),
```
veganstraightedge:
```suggestion
                   image_tag(url_for(article.image), class: "u-photo", alt: article.image_description),
```

app/views/articles/_image_media.html.erb
```diff
@@ -1,7 +1,19 @@
...
+        article.path %>
```
veganstraightedge:
```suggestion
                   article.path %>
```

config/storage.yml
veganstraightedge:
```suggestion
  bucket: <%= ENV['AWS_BUCKET'] %>
```

config/storage.yml
```diff
@@ -6,29 +6,17 @@ local:
...
 
-# Use rails credentials:edit to set the AWS secrets (as aws:access_key_id|secret_access_key)
+amazon:
+  service: S3
+  access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
+  secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
+  region: 'us-west-2'
+  bucket: <%= ENV['AWS_BUCKET'] %>
+  
+# TODO: Use rails credentials:edit to set the AWS secrets (as aws:access_key_id|secret_access_key)
```
astronaut-wannabe: @veganstraightedge the rails docs suggest doing this, but i left the ENV var version above for now. lemme know if you want to change to this credentials api instead

veganstraightedge:
We’ve rethought how we wanna approach uploads. So, we’re closing this PR to restart it afresh. The PR changes will still be available for reference.

PR #855, opened 2018-09-24
 Make bugsnag differentiate staging and production 

astronaut-wannabe:
This is to make bugsnag notifications easier to
differentiate


astronaut-wannabe:
@veganstraightedge I think this will work to create a staging rails environment that is the same as production (but has `staging` name in bugsnag)

I think the last piece is setting `RAILS_ENV=staging` instead of `production` on the rails staging hosts. can you do that part and give it a test?

PR #876, opened 2018-10-21
Set locale from Accept-Language header and subdomain

nornagon:
{es,en,...}.crimethinc.com, if present, will override the Accept-Language header.


app/controllers/application_controller.rb
```diff
@@ -50,8 +50,11 @@ def creating?
   end
   helper_method :creating?
 
-  def set_locale
-    I18n.locale = I18n.default_locale
+  def subdomain_locale_override
+    locale = request.subdomain
+    if I18n.available_locales.include?(locale.to_sym)
+      I18n.locale = locale
+    end
```
veganstraightedge:
```suggestion
    else
      I18n.locale = I18n.default_locale
    end
```

veganstraightedge: I think you still need to fall back to setting the default locale in the case where the requested locale / subdomain isn't in our list of available locales.

app/controllers/application_controller.rb
```diff
@@ -50,8 +50,14 @@ def creating?
...
-  def set_locale
-    I18n.locale = I18n.default_locale
+  def subdomain_locale_override
+    locale = request.subdomain
+    I18n.locale = locale if I18n.available_locales.include?(locale.to_sym)
+
+    # Force the subdomain to match the locale.
+    # Don't do this in development, because typically local development
+    # environments don't support subdomains (en.localhost doesn't resolve).
+    return redirect_to subdomain: I18n.locale if I18n.locale != I18n.default_locale and request.subdomain.empty? && Rails.env.production?
```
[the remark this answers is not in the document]
nornagon: nice! suggestion feature is great :) rubocop is mean tho

PR #886, opened 2018-10-27
Add simple language switcher to nav

veganstraightedge:
This adds a simple language switcher to the header navbar. It will only work for a short list of support languages. The redesign has a proper language switcher in the nav.

**DO NOT MERGE** into production yet.


veganstraightedge:
Replaced by https://github.com/crimethinc/website/pull/955

Issue #541, opened 2017-12-04
Add UI for publication status to Tools and Podcast Episodes


veganstraightedge:
`status_id` has been added to more resources other than just `Article`, but only the `Article` form has UI for setting the `Status` as `draft` or `published`. 

This issue is to add that UI to the other forms too. And preferably refactor all of them into one partial or something.

- [ ] Article
- [ ] Book / Zine
- [ ] Poster / Sticker
- [ ] Logo
- [ ] Video
- [ ] DRY refactor into one shared partial
- [ ] Ensure the JS to `Publish NOW` from Article#form works on all above forms

PR #893, opened 2018-10-29
Add status UI to book/zines

astronaut-wannabe:
The changes to `publication_status_badge_class` might not be
necessary, but when testing with the books loaded from the `seed`
script I got errors because all of the books had a status of `nil`

I think they all have a status in the prod db and maybe I should just
update the `seed` script to set `published` to all of the books.

thoughts?

# screenshots
### index
screen shot 2018-10-28 at 5 49 01 pm
### form
screen shot 2018-10-28 at 5 38 19 pm

---

### index
screen shot 2018-10-28 at 6 14 52 pm

### form
screen shot 2018-10-28 at 5 48 51 pm

relates to #541


astronaut-wannabe:
@veganstraightedge ok, will update the seeds and revert the `'??????'` change

PR #895, opened 2018-10-29
Hardcode available_locales to english and spanish, for now

veganstraightedge:
To prevent other language failing before we have translations for them.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -30,7 +30,7 @@ module Crimethinc
     config.i18n.default_locale = :en
 
     # Whitelist locales available for the application
-    config.i18n.available_locales = [:en, :es]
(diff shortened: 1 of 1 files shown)


veganstraightedge:
@guitsaru I **think** it'll only redirect to a subdomain that is:

- in this list AND
- is set by your browser as your preferred language

guitsaru:
@veganstraightedge right, isn't that the case for pt-BR that we're trying to fix?  I just want to make sure that doesn't keep happenning.

astronaut-wannabe:
should be ok since browsers send actual locale code (e.g. `pt-BR` whereas TCE is in a word)

guitsaru:
@astronaut-wannabe gotcha, I thought that might be the case, but wanted to make sure.

PR #910, opened 2018-11-03
Denormalize Status into #publication_status on objects

### LINKED ISSUE #858: Denormalize Status on Articles/etc
Originally, it was imagined that there would be multiple stages of the publishing flow, like in WordPress:

- draft
- edited
- designed
- published

In practice, there has only been two `draft` and `published`.

Because we have a `Status` model/table, every request has one/many extra database lookups for the publication status of things.

We _could_ do away with `Status` altogether and store the published bit on the Articles/etc themselves.

`published_status: draft`.

We could even go as far as only storing the `published_at` timestamp.

- If `published_at` is nil/blank, the thing is `draft`
- If `published_at` is before Time.now, the thing is `published` and `live`
- If `published_at` is after Time.now, the thing is `published`, but not `live`

Or something like that…

veganstraightedge:
Closes #858.
Relates to #541.

***

This replaces all of the uses of `Status`, `.status`, and `status_id` with `publication_status` as a String.

`publication_status: 'draft'`
or
`publication_status: 'published'`


app/controllers/admin/books_controller.rb
```diff
@@ -66,17 +63,12 @@ def set_publication_type
...
-      @published = Status.find_by(name: 'published')
-    end
-
     def set_ebook_formats
       @ebook_formats = Tool::EBOOK_FORMATS
     end
 
     def book_params
-      params.require(:book).permit(:title, :subtitle, :content, :tweet, :summary, :status_id,
+      params.require(:book).permit(:title, :subtitle, :content, :tweet, :summary, :status_id, :publication_status,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/journals_controller.rb
```diff
@@ -74,7 +74,8 @@ def journal_params
                                       :lite_download_present, :epub_download_present, :mobi_download_present,
                                       :print_black_and_white_a4_download_present, :print_color_a4_download_present,
                                       :print_color_download_present, :print_black_and_white_download_present,
-                                      :screen_single_page_view_download_present, :screen_two_page_view_download_present)
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/logos_controller.rb
```diff
@@ -53,15 +50,11 @@ def set_logo
...
-    def set_statuses
-      @draft     = Status.find_by(name: 'draft')
-      @published = Status.find_by(name: 'published')
-    end
-
     def logo_params
       params.require(:logo).permit(:title, :subtitle, :description,
-                                   :slug, :height, :width, :content_format, :summary, :status_id, :published_at,
-                                   :jpg_url_present, :png_url_present, :pdf_url_present, :svg_url_present, :tif_url_present)
+                                   :slug, :height, :width, :content_format, :summary, :status_id,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/pages_controller.rb
```diff
@@ -76,10 +70,9 @@ def organize_page
     end
 
     def page_params
-      params.require(:page).permit(:year, :month, :day, :css, :slug, :tags,
-                                   :draft_code, :status_id, :published_at,
-                                   :categories, :hide_header, :hide_footer,
-                                   :hide_layout, :published_at_tz)
+      params.require(:page).permit(:year, :month, :day, :css, :slug, :tags, :draft_code, :status_id,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/posters_controller.rb
```diff
@@ -65,20 +62,16 @@ def set_publication_type
...
-    end
-
     def poster_params
       params.require(:poster).permit(:sticker, :title, :subtitle, :content, :content_format,
-                                     :buy_info, :buy_url, :price_in_cents, :summary, :description, :published_at, :status_id,
-                                     :front_image_present, :back_image_present, :front_download_present, :back_download_present,
-                                     :slug, :height, :width, :depth, :front_image_format, :back_image_format,
-                                     :front_color_image_present, :front_black_and_white_image_present, :back_color_image_present,
+                                     :buy_info, :buy_url, :price_in_cents, :summary, :description,
+                                     :published_at, :status_id, :front_image_present, :back_image_present,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/stickers_controller.rb
```diff
@@ -65,15 +62,12 @@ def set_publication_type
...
-      @draft     = Status.find_by(name: 'draft')
-      @published = Status.find_by(name: 'published')
-    end
-
     def poster_params
       params.require(:poster).permit(:sticker, :title, :subtitle, :content, :published_at,
-                                     :content_format, :buy_info, :buy_url, :price_in_cents, :summary, :description, :status_id,
-                                     :front_image_present, :back_image_present, :lite_download_present, :slug, :height, :width)
+                                     :content_format, :buy_info, :buy_url, :price_in_cents,
+                                     :summary, :description, :status_id, :front_image_present,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/videos_controller.rb
```diff
@@ -61,16 +58,11 @@ def set_video
       @video = Video.find(params[:id])
     end
 
-    def set_statuses
-      @draft     = Status.find_by(name: 'draft')
-      @published = Status.find_by(name: 'published')
-    end
-
     def video_params
       params.require(:video).permit(:title, :subtitle, :status_id,
```
astronaut-wannabe: remove `status_id`

app/controllers/admin/zines_controller.rb
```diff
@@ -86,7 +78,8 @@ def book_params
                                    :lite_download_present, :epub_download_present, :mobi_download_present,
```
astronaut-wannabe: remove `status_id`

db/schema.rb
```diff
@@ -13,7 +13,6 @@
 ActiveRecord::Schema.define(version: 2018_11_03_020911) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
astronaut-wannabe: these just look like noise.

will there be a migration to drop the status table after this change?

app/models/publication_status.rb
```diff
@@ -0,0 +1,3 @@
+class PublicationStatus
```
astronaut-wannabe: interested why a model instead of an enum on the publishable concern (e.g. [this][0])

[0]:https://api.rubyonrails.org/v5.2.1/classes/ActiveRecord/Enum.html

app/models/concerns/slug.rb
```diff
@@ -13,7 +13,7 @@ def slug_exists?
   def blank_slug
     return unless is_a?(Article) || is_a?(Page)
 
-    self.slug = nil if status.blank? || (status.present? && !published?)
+    self.slug = nil if publication_status.blank? || (publication_status.present? && !published?)
```
astronaut-wannabe: when we add the column can we make the defaultt value `draft` and then have a nonnull restraint. that way gauranteed not to be blank

db/schema.rb
```diff
@@ -90,13 +89,13 @@
...
     t.boolean "mobi_download_present"
+    t.integer "status_id"
     t.boolean "print_black_and_white_a4_download_present"
     t.boolean "print_color_a4_download_present"
     t.boolean "print_color_download_present"
     t.boolean "print_black_and_white_download_present"
     t.boolean "screen_single_page_view_download_present"
     t.boolean "screen_two_page_view_download_present"
-    t.integer "status_id"
     t.string "publication_status", default: "draft"
```
astronaut-wannabe: can `publication_status` be `null: false`per my comment above

db/schema.rb
veganstraightedge: Yeah. I plan to do that as a separate PR. Just in case anything goes wrong, we'll still have the `statuses` table and all of the `status_id` references.

app/models/concerns/slug.rb
veganstraightedge: Great idea!

app/models/publication_status.rb
veganstraightedge: Because I didn't think of that last night. 🤦🏻‍♀️

I'll change that.

db/schema.rb
veganstraightedge: Yep 👍🏻

app/controllers/admin/books_controller.rb
veganstraightedge: Did all of these. 👍🏻

app/models/publication_status.rb
veganstraightedge: Removed this class and moved to an enum on Publishable. Thanks for the suggestion!

app/models/concerns/publishable.rb
```diff
@@ -1,12 +1,13 @@
...
+  PUBLICATION_STATUSES = %i[draft published].freeze
   extend ActiveSupport::Concern
 
   included do
     belongs_to :status
+    enum publication_status: PUBLICATION_STATUSES
+
     default_scope { order(published_at: :desc) }
 
-    scope :draft,         -> { where(status: Status.find_by(name: 'draft')) }
```
astronaut-wannabe: nice! i didnt knojw we'd get these built-in with an enum

app/views/admin/_published_at.html.erb
```diff
@@ -1,5 +1,5 @@
 <div id="datetime" class="py-5">
-  <% if resource.new_record? || (resource.status.present? && !resource.published?) %>
+  <% if resource.draft? %>
```
astronaut-wannabe: digging these `if draft?` changes. way clearer on a quick read-through

app/models/concerns/publishable.rb
astronaut-wannabe: nice! i didnt knojw we'd get these built-in with an enum

db/migrate/20181104000207_change_publication_status_to_not_null_integer.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    remove_column :articles, :publication_status, :string, default: 'draft'
+    remove_column :books,    :publication_status, :string, default: 'draft'
+    remove_column :journals, :publication_status, :string, default: 'draft'
+    remove_column :logos,    :publication_status, :string, default: 'draft'
+    remove_column :pages,    :publication_status, :string, default: 'draft'
+    remove_column :posters,  :publication_status, :string, default: 'draft'
+    remove_column :videos,   :publication_status, :string, default: 'draft'
+
+    add_column :articles, :publication_status, :integer, default: 'draft', null: false
```
astronaut-wannabe: actually, let me think through this migration a bit...ui think it might make everything `draft` on deploy

PR #911, opened 2018-11-03
[WIP] add espanolamericalatina TCE page

astronaut-wannabe:


config/initializers/locale.rb
```diff
@@ -0,0 +1,14 @@
...
+I18n.default_locale = :en
+
+# Whitelist locales available for the application
+I18n.available_locales = %i[espanolamericalatina en es lietuvos 한국어 english espanol فارسی]
+
+# Allow nested diretories in locales
+I18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
+
+# custom TCE fallbacks
+I18n.fallbacks.map(espanolamericalatina: :espanol)
```
astronaut-wannabe: this is supposed to work according to the i18n gem, but it isn't working for me locally

astronaut-wannabe: i gave up trying to do fallbacks. think it wasn't working because of the non-standard way i am using I18n in TCE

astronaut-wannabe:
@veganstraightedge I didnt see any missing Ⓐs on a wecond proofread, but i wouldht be suprised if there are some

I will add fixes if any bug reports come in

PR #912, opened 2018-11-03
Text Only Site

GeoffFoley:
Adds code for a text-only version of the site.

# #480

### DIFF
--- a/app/assets/stylesheets/lite_mode.scss
+++ b/app/assets/stylesheets/lite_mode.scss
@@ -1,0 +1,156 @@
+body {
+  font-family: system, -apple-system, BlinkMacSystemFont, ‘Segoe UI’, Roboto, Oxygen, Ubuntu, Cantarell, ‘Open Sans’, ‘Helvetica Neue’, sans-serif;
+
+  img {
+    display: none;
+  }
+
+  &#products {
+    .product {
+      margin-top: 1rem;
+      border-top: 1px solid #ddd;
+      padding-top: 1rem;
+
+      .product-front {
+        > a {
+          display: none;
+        }
+      }
+
+      .product-main {
+        > header {
+          display: none;
+        }
+      }
+    }
+
+    .row#extra {
+      border-top: 1px solid #ddd;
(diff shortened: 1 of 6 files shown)


GeoffFoley:
Come on TCI *fingers crossed*

GeoffFoley:
And now I will remember ./script/test *sigh*

app/assets/stylesheets/lite_mode.scss
```diff
@@ -0,0 +1,156 @@
...
+
+      .product-front {
+        > a {
+          display: none;
+        }
+      }
+
+      .product-main {
+        > header {
+          display: none;
```
veganstraightedge: Why is this being hidden? It seems important.

For example, this zine page:

`https://crimethinc-staging-pr-912.herokuapp.com/zines/secret-world-of-duvbo`

It doesn't have a title/subtitle and the nav crumb trail. 

Compared to:

`https://crimethinc.com/zines/secret-world-of-duvbo`

veganstraightedge:
This is really great! And pretty close to ready to go.

I think the only two changes that this PR needs are:

- [ ] DRY up the `Article` card background image conditional. 
https://github.com/crimethinc/website/pull/912#discussion_r230570277
- [ ] Remove the `display: none` on products to re-show the title/subtitles on the various `Tools`

Oh, and removing the `def lite_mode?; true ;end`, of course

After that, I think it's shippable!

PS. The `lite.` subdomain is already hooked up.

app/assets/stylesheets/lite_mode.scss
GeoffFoley: I hid that because if not, the titles are doubled up on the books page. I will try and do some more specific targeting for the books page. No zines show up on the Zine page in my dev env so I didn't see that the classes were reused on those.

veganstraightedge:
Great job, @GeoffFoley! 

There are some polish things to do, but I'll write them up as separate issues to keep this PR from scope creeping. 

But… :shipit: ‼️

Issue #916, opened 2018-11-05
Inline CSS into style block on Lite Mode


veganstraightedge:
When on `lite.crimethinc.com`, I think we should stick the lite mode CSS into a `<style>`
block in the `<head>`, since it's such a small amount of CSS and to reduce the HTTP requests by one.

PR #954, opened 2018-11-23
Inline Lite Mode CSS

GeoffFoley:
Inline CSS into style block on Lite Mode #916


app/views/shared/head/_lite_mode_css.erb
```diff
@@ -0,0 +1,148 @@
+
+<!-- Lite Mode Styles -->
+<style>
+  @charset "UTF-8";
```
veganstraightedge: I had never seen this before, so I looked it up. And as I read this selection from MDN, it seems like it doesn't get applied from within a `<style>` block on a page. Instead, I'm guessing, the page's charset gets used for the CSS in the `<style>` block.

Again, I'm new to this line of code, so maybe there's something that I don't know about / am misunderstanding.

> The @charset CSS at-rule specifies the character encoding used in the style sheet. It must be the first element in the style sheet and not be preceded by any character; as it is not a nested statement, it cannot be used inside conditional group at-rules. If several @charset at-rules are defined, only the first one is used, and **it cannot be used inside a style attribute on an HTML element or inside the <style> element** where the character set of the HTML page is relevant.

https://developer.mozilla.org/en-US/docs/Web/CSS/@charset

app/views/shared/head/_lite_mode_css.erb
```diff
@@ -0,0 +1,148 @@
...
+
+  .p-x-subtitle {
+    margin-top: 0;
+    font-size: 1.25em;
+  }
+
+  img[src*="start-header"] {
+    display: none;
+  }
+</style>
```
veganstraightedge: Please rename the file to `_lite_mode_css.html.erb` since it is technically a snippet of HTML because of the `<style>` tags.

app/views/shared/head/_lite_mode_css.erb
GeoffFoley: OK. I'll remove. I had just straight copied the rendered CSS from the compiled (sass) source. Something in the rails sass compilation is adding that in the compiled css file.

veganstraightedge: Oh! Good to know. 👍🏻

PR #971, opened 2018-11-26
Remove unneeded explicit render to the view that matches the controller/action

veganstraightedge:


config/locales/pages/de/about.markdown
```diff
@@ -0,0 +1,60 @@
...
+**CrimethInc. is a desperate venture.** As this society lurches ever closer to annihilation, we are staking everything on the possibility that we could kick open the escape hatch to another future. Rather than competing for social capital or selling ourselves to the highest bidder, we have thrown ourselves completely into the struggle for a better world. We invite you to do the same.
+
+## General Inquiries
+
+_Email:_ \\
+[hello@crimethinc.com](mailto:hello@crimethinc.com)
+
+_Twitter:_ \\
+[@crimethinc](https://twitter.com/crimethinc)
+
```
just1602: @veganstraightedge should we put the fb page here since it is active and respond pretty fast.

just1602: @veganstraightedge should we put the fb page here since it is active and respond pretty fast.

veganstraightedge: That was an accidental addition to this branch.

PR #976, opened 2018-11-28
[TCE] Update TCE generator to add locale to config

astronaut-wannabe:
We now explicitly set supported locales in `application.rb`, add this
step to the generator so it is no longer manual


astronaut-wannabe:
@mofo37 this is to get rid of the manual step i mentioned earlier

PR #992, opened 2018-12-14
Fix locale site-switcher

astronaut-wannabe:
If you cache on a `class` method, then it is cached as long as that
`class` is in memory

### DIFF
--- a/app/models/locales.rb
+++ b/app/models/locales.rb
@@ -3,13 +3,7 @@ class Locales
 
   class << self
     def supported
-      @supported ||= supported_locales
-    end
-
-    private
-
-    def supported_locales
-      @supported = SUPPORTED - [I18n.locale]
+      SUPPORTED - [I18n.locale]
     end
   end
 end


app/models/locales.rb
```diff
@@ -3,13 +3,7 @@ class Locales
...
   class << self
     def supported
-      @supported ||= supported_locales
-    end
-
-    private
-
-    def supported_locales
-      @supported = SUPPORTED - [I18n.locale]
+      SUPPORTED - [I18n.locale]
```
astronaut-wannabe: just to clarify what this is intendinng to do, and what it was doing:

## intention
this code wants to show all of the site locales that are supported *EXCEPT* for whatever locale the user currently has (`I18n.locale`)

so, 
- for me: `[de en es] - [en] = [de es]`
- for a browser in german: `[de en es] - [de] = [en es]`

## what it was actually doing
this calculation needs to be run for *every request* (since different people will have different `I18n.locale`s set.

However, it was only being run once (per app host), and then memoized (cached in memory).

Since this is on a static class, the cache lives as long as the class is in memory (and since it runs every request this class will probably never be garbage collected)

This affectivley means the function will only ever return the value from the very first request that triggers it

Issue #960, opened 2018-11-24
Podcasts in Lite Mode


veganstraightedge:
Don't render images and audio.

## Examples
https://lite.crimethinc.com/podcast
https://lite.crimethinc.com/podcast/hotwire-46

## TODO
- [x] /podcast (`podcast#index`)
  - [x] podcast cover image
  - [x] latest episode audio
- [x] /podcast/:episode (`podcast#show`)
  - [x] podcast cover image
  - [x] episode audio

***

Give the subscribe links some breathing room.

- [x] subscribe links CSS, maybe separate onto their own lines?
- <img width="375" alt="screen shot 2018-11-23 at 6 41 41 pm" src="https://user-images.githubusercontent.com/4361/48963801-9bca9580-ef4f-11e8-8b4c-2778110ac3f3.png">

GeoffFoley:
Closed with #1052

PR #1052, opened 2019-02-25
Podcasts in Lite Mode #960

GeoffFoley:
Hides the podcast player on the Podcast index and show pages, as well as the "Listen..." text on the show page, if in Lite Mode.

### DIFF
--- a/app/views/podcast/show.html.erb
+++ b/app/views/podcast/show.html.erb
@@ -16,18 +16,20 @@
 
     <div class="column column-two-third">
       <div class="e-content">
-
+				
         <h3 class="listen">
-          <%= t("podcast.listen") %>
-          <% if @episode.duration.present? %>
-            <small>
-              —
-              <%= @episode.duration %>
-              <abbr title="minutes">min</abbr>
-            </small>
+          <% if !lite_mode? %>
+          	<%= t("podcast.listen") %>
+          	<% if @episode.duration.present? %>
+            	<small>
+              	—
+              	<%= @episode.duration %>
+            		<abbr title="minutes">min</abbr>
+          		</small>
+        		<% end %>
           <% end %>
         </h3>
-        <%= render "podcast/player", episode: @episode %>
+        <%= render "podcast/player", episode: @episode unless lite_mode? %>
 
         <div class="episode-info">
           <div class="buttons">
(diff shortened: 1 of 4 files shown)


app/views/podcast/show.html.erb
```diff
@@ -16,18 +16,20 @@
 
     <div class="column column-two-third">
       <div class="e-content">
-
+
```
veganstraightedge: Could you delete this trailing blank space please?

app/views/shared/head/_lite_mode_css.html.erb
```diff
@@ -14,6 +14,14 @@
...
     display: none;
   }
+	
+	body#podcast .subscribe a.u-url {
+		display: block;
+	}
+	
+	body#podcast .subscribe a.u-url + br {
+		display: none;
+	}
```
veganstraightedge: It looks like maybe your editor is set to use tabs instead of spaces. Would you mind find/replacing those tabs with two spaces so that these new CSS rules line up with the others, please?

app/views/podcast/show.html.erb
```diff
@@ -16,18 +16,20 @@
...
-            </small>
+          <% if !lite_mode? %>
+          	<%= t("podcast.listen") %>
+          	<% if @episode.duration.present? %>
+            	<small>
+              	—
+              	<%= @episode.duration %>
+            		<abbr title="minutes">min</abbr>
+          		</small>
+        		<% end %>
```
veganstraightedge: This might be a tabs/spaces thing too (see below), but would you mind lining up/nesting the lines so that match up, please?

```
if
  small
    -
    ep.dur
    abbr /abbr
  /small
end
```

app/views/shared/head/_lite_mode_css.html.erb
```diff
@@ -14,6 +14,14 @@
   body img {
     display: none;
   }
+	
+	body#podcast .subscribe a.u-url {
+		display: block;
+	}
+	
+	body#podcast .subscribe a.u-url + br {
+		display: none;
```
veganstraightedge: I'm curious why this is done in CSS and not as another `if lite_mode?` in the view. What was your thinking on that?

GeoffFoley: I thought that through css would be a cleaner way to hide the br tag as opposed to logic in the view. I can do it the other way, if that's what you'd prefer.

veganstraightedge: Either way. I was just curious. Let's go with the one that you think is best!

PR #1066, opened 2019-03-06
WIP: Delete Status from most places

veganstraightedge:
The `enum` of `publication_status` has replaced all associations to the model `Status`. It turns out that we only ever use `draft` or `published`. No need for the extra SQL query on all of the things.

This PR removes the final unused traces of `Status`.

***

# TODO

- [x] Remove `status_id` from the Scenic views
- [x] Add the two migrations from below
- [x] Migrate
- [x] Confirm that tests are green still
- [x] Spot check that the site behaves properly


veganstraightedge:
`20190306082914_drop_statuses.rb`

```ruby
class DropStatuses < ActiveRecord::Migration[5.2]
  def up
    drop_table :statuses
  end

  def down
    create_table :statuses do |t|
      t.string :name

      t.timestamps
    end
  end
end

```

veganstraightedge:
`20190306083154_remove_status_id_columns.rb`

```ruby
class RemoveStatusIdColumns < ActiveRecord::Migration[5.2]
  def change
    remove_column :articles, :status_id, :integer
    remove_column books, :status_id, :integer
    remove_column journals, :status_id, :integer
    remove_column logos, :status_id, :integer
    remove_column pages, :status_id, :integer
    remove_column posters, :status_id, :integer
    remove_column stickers, :status_id, :integer
    remove_column videos, :status_id, :integer
    remove_column zines, :status_id, :integer
  end
end
```

PR #1094, opened 2019-04-03
Podcast views refactoring

just1602:
So far, I move the episode related stuff to the new `EpisodesController`. Now I've to find a way to make the url go from `/podcast/:slug` to `/podcast/:podcast_slug/ episodes/:id` or `/podcast/:podcast_slug/episodes/:id` and then create the podcast show view.

- [x] move episode stuff to the `EpisodesController`
- [x] create the routes pattern to make episode as a podcast sub resource
- [x] create the detail podcast view that contains a podcast info and all its episodes


just1602:
@veganstraightedge could you check this when you have 2 seconds? I'm not sure if we'll be able to make redirects from old to new url the way I'm going.

config/routes.rb
```diff
@@ -64,10 +64,11 @@
...
   # Podcast
-  get 'podcast/feed',             to: 'podcast#feed',       as: :podcast_feed, defaults: { format: 'rss' }
-  get 'podcast',                  to: 'podcast#index',      as: :podcast
-  get 'podcast/:slug',            to: 'podcast#show',       as: :episode
-  get 'podcast/:slug/transcript', to: 'podcast#transcript', as: :episode_transcript
+  get 'podcast/feed',             to: 'podcast#feed',        as: :podcast_feed, defaults: { format: 'rss' }
+  get 'podcast',                  to: 'podcast#index',       as: :podcast
+  get 'podcast/:slug',            to: 'podcast#show',        as: :podcast_detail
+  get 'podcast/:slug/episodes/:id', to: 'episodes#show',     as: :podcast_episode
+  get 'podcast/:slug/transcript', to: 'episodes#transcript', as: :episode_transcript
```
veganstraightedge:
```suggestion
  get 'podcast/:slug/episodes/:id/transcript', to: 'episodes#transcript', as: :episode_transcript
```

A transcript needs to hang off of a single episode.

app/controllers/episodes_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @editable = @episode
+    @title    = title_for :podcasts, @episode.name, :transcript
+
+    render 'podcast/show'
+  end
+
+  private
+
+  def set_episode
+    @episode = Episode.where(id: params[:id])
```
veganstraightedge: It seem like to me that you'll also need to include the `podcast_slug` in the lookup here, since the URL will be `podcast/:slug/episodes/:id`

And there will be multiple episode `1`. One for each podcast.

just1602: Should I had something like `podcast_episode_id` ? Because right now, there is only the slug to work with and the id is the episode model id?

veganstraightedge: That sounds like a good idea. 👍🏻

config/routes.rb
```diff
@@ -74,9 +74,9 @@
 
   # Podcast
   get 'podcast/feed',                           to: 'podcast#feed',        as: :podcast_feed, defaults: { format: 'rss' }
-  get 'podcasts',                               to: 'podcast#index',       as: :podcast
-  get 'podcasts/:slug',                         to: 'podcast#show',        as: :podcast_detail
-  get 'podcasts/:slug/episodes/:id',            to: 'episodes#show',       as: :podcast_episode
+  get 'podcasts',                               to: 'podcast#index',       as: :podcasts
```
just1602: Maybe in a future PR we should think about use `resources` and `resource` and make the routes as restfull as possible?

app/controllers/episodes_controller.rb
veganstraightedge: I ended up adding `episode_number` to `Episode`, because it's not quite actually an `id`. But same general idea.

config/routes.rb
veganstraightedge: Totally!

veganstraightedge:
Here is the list of redirects that I plan to add to production when we merge this PR into production.

https://gist.github.com/veganstraightedge/3bb258a582dc214516095d0426c1b7db

app/models/podcast.rb
```diff
@@ -5,7 +5,7 @@ class Podcast < ApplicationRecord
   has_many :episodes, dependent: :destroy
 
   def path
-    '/podcast'
+    "/podcasts/#{slug}"
```
guitsaru: Is `slug` a required value?  What happens if slug is blank?

veganstraightedge: That's a good point / question. 👍🏻

I think it's not currently required by code, but has been by social practice. We should codify it though. Because weird shit happens without one.

We can do that in a separate PR.
All current podcasts have slugs and new podcasts don't come around that often.

PR #1162, opened 2019-06-11
Add spec to lock in current Pages behavior

astronaut-wannabe:
This PR adds a system spec that verifies the current behavior of the
`Page`s we have in production

Current Behavior
---------------------
Currently in production we have the following pages
```ruby
#=> [“/about”, “/start”, “/faq”, “/contact”]
```

Also, currently, if a page doesn't exist we currently re-direct to the
homepage instead of throwing a `404`

Why?
-------
We want to deprecate the `Page` feature (it isn't really used) and get
rid of the catchall redirect so that we start showing `404`s again.

The main two reasons for showing `404`s again are:
1. we are pretty stable now, and would be nice to fix broken links in the wild with redirects
2. this should make integrating with `ActiveStorage` easier as there
   is currently some funky behavior with the catchall route

This PR Does Not
----------------
Note that this PR doesn't change any of the current behavior (leaves
the catchall route in place)

This PR just locks in a spec for the current content so that when the
route file is changed in a future PR, we don't regress any of these pages

### DIFF
--- a/spec/system/special_pages_spec.rb
+++ b/spec/system/special_pages_spec.rb
@@ -39,7 +39,7 @@ describe 'Navigating to special 1-off pages' do
   end
 
   context 'with a non-existing page path' do
-    it 'redirect to thee home page' do
+    it 'redirect to the home page' do
       visit '/blahblahblah'
       expect(page).to have_current_path '/'
     end
--- a/spec/system/special_pages_spec.rb
+++ b/spec/system/special_pages_spec.rb
@@ -1,0 +1,81 @@
+require 'rails_helper'
+
+describe 'Navigating to special 1-off pages' do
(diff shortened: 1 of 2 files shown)


spec/system/special_pages_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      visit '/faq'
+      expect(page).to have_current_path('/2016/09/28/feature-the-secret-is-to-be' \
+                                        'gin-getting-started-further-resources-f' \
+                                        'requently-asked-questions')
+      expect(page).to have_content 'Frequently Asked Questions about Anarchism'
+    end
+  end
+
+  context 'with a non-existing page path' do
+    it 'redirect to thee home page' do
```
astronaut-wannabe:
```suggestion
    it 'redirect to the home page' do
```

PR #1165, opened 2019-06-12
modify scenic search view

astronaut-wannabe:
@veganstraightedge @guitsaru I followed the scenic docs to [update  a view][0]

the main change i made to the view is i deleted this `JOIN`:
```sql
INNER JOIN statuses ON statuses.id = articles.status_id AND statuses.name='published'
```

and then I modified the `WHERE` clause to check for published articles/pages:

```sql
WHERE pages.published_at < now() AND pages.publication_status = 1
```

~the last thing to do was run `rails db:migrate`, but I got this error:~

```sh
ActiveRecord::StatementInvalid: PG::WrongObjectType: ERROR:  "search_results" is not a view
```
~not sure if that is just due to my local DB not having the scenic views in it. thoughts??~

@guitsaru fixed it. needed to add `materialized: true` to the migration 

[0]:https://github.com/scenic-views/scenic#cool-but-what-if-i-need-to-change-that-view

### DIFF
--- a/db/views/search_results_v05.sql
+++ b/db/views/search_results_v05.sql
@@ -24,7 +24,7 @@ SELECT
     LEFT JOIN tags ON tags.id = taggings.tag_id
     LEFT JOIN categorizations ON categorizations.article_id = articles.id
     LEFT JOIN categories ON categories.id = categorizations.category_id
-    WHERE articles.published_at < now() AND articles.   publication_status = 1
+    WHERE articles.published_at < now() AND articles.publication_status = 1
     GROUP BY searchable_id, searchable_type
 
     UNION
--- a/db/views/search_results_v05.sql
+++ b/db/views/search_results_v05.sql
@@ -1,0 +1,58 @@
+SELECT
+  searchable_id, searchable_type, title, subtitle, content,
+  tag_names AS tag,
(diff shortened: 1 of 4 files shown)


db/views/search_results_v05.sql
```diff
@@ -0,0 +1,58 @@
...
+      to_tsvector(coalesce(articles.content, '')) AS content,
+      CASE WHEN COUNT(tags) = 0 THEN ARRAY[]::text[] ELSE array_remove(array_agg(tags.name), NULL) END AS tag_names,
+      CASE WHEN COUNT(categories) = 0 THEN ARRAY[]::text[] ELSE array_remove(array_agg(categories.name), NULL) END AS category_names
+
+    FROM articles
+    LEFT JOIN taggings ON taggings.taggable_id = articles.id AND taggings.taggable_type = 'Article'
+    LEFT JOIN tags ON tags.id = taggings.tag_id
+    LEFT JOIN categorizations ON categorizations.article_id = articles.id
+    LEFT JOIN categories ON categories.id = categorizations.category_id
+    WHERE articles.published_at < now() AND articles.   publication_status = 1
```
guitsaru: `articles.   publication_status` extra space here.

PR #1221, opened 2019-07-31
[Translations] Allow article locale to override rtl/ltr html attr

astronaut-wannabe:
So, it looks like articles can have on-the-fly locales created (this
is separate from having site-wide support for a language)

The mechanism added to do this does not account for `rtl` languages.

This changes adds some (maybe janky) support for `rtl`

### NOTE: an article set `rtl` will show the `admin#edit` page in
RTL (which looks _okay_, but some CSS should probably be re-written to
make it look _good_)

### DIFF
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -1,10 +1,10 @@
 module TagsHelper
   def html_dir
-    t 'language_direction'
+    article_locale&.language_direction.presence || t('language_direction')
   end
 
   def html_lang
-    I18n.locale
+    article_locale&.abbreviation.presence || I18n.locale
   end
 
   def html_prefix
@@ -22,5 +22,11 @@ module TagsHelper
 
   def body_id
     @body_id
+  end
+
+  private
+
+  def article_locale
+    @article_locale ||= Locale.find_by(abbreviation: @article&.locale)
   end
 end
--- a/app/views/shared/footer/_newsletter.html.erb
+++ b/app/views/shared/footer/_newsletter.html.erb
@@ -27,7 +27,8 @@
         <%= button_tag t("footer.contact.newsletter.signup_button_text"), name: "subscribe", id: "mc-embedded-subscribe" %>
 
         <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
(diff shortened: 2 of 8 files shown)


app/helpers/tags_helper.rb
```diff
@@ -1,10 +1,18 @@
...
+      article_locale.language_direction
+    else
+      t 'language_direction'
+    end
   end
 
   def html_lang
-    I18n.locale
+    if article_locale.present?
+      article_locale.language_direction
```
astronaut-wannabe:
```suggestion
      article_locale.abbreviation
```

app/helpers/tags_helper.rb
```diff
@@ -1,10 +1,18 @@
 module TagsHelper
   def html_dir
-    t 'language_direction'
+    if article_locale.present?
+      article_locale.language_direction
+    else
+      t 'language_direction'
+    end
```
veganstraightedge: Could you use `.presence` here?

```ruby
article_locale.language_direction.presence || t 'language_direction'
```

app/helpers/tags_helper.rb
```diff
@@ -23,4 +31,10 @@ def html_class
   def body_id
     @body_id
   end
+
+  private
+
+  def article_locale
+    @article_locale ||= Locale.find_by(abbreviation: @article&.locale)
```
veganstraightedge: Will this memoization persist past this request/article? I feel like we had locale / cache issues before.

astronaut-wannabe: no, should only persist across page renders (the issue before was due to modifying `I18n.locale` directly)

app/views/shared/footer/_newsletter.html.erb
```diff
@@ -27,7 +27,8 @@
         <%= button_tag t("footer.contact.newsletter.signup_button_text"), name: "subscribe", id: "mc-embedded-subscribe" %>
 
         <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
-        <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_cb48383602b7373f496e9ba33_344714b249" tabindex="-1" value=""></div>
+        <% left_or_right = html_dir == 'ltr' ? "position: absolute; left: -5000px;" : "position: absolute; right: -5000px;" %>
```
veganstraightedge: Good catch!

PR #1227, opened 2019-08-03
remove action-cable

astronaut-wannabe:
this part of rails is not currently used in our app

I left the Gemfile as

```ruby
  gem 'rails'
```

and the `config/application.rb` as

```ruby
require 'rails/all'
```

to make it easier to add back in case we use it in the future, but
this just removes all of the javascript files and directories so
people do not think they are being used for something


astronaut-wannabe:
well, this doesn't look right: https://crimethinc-staging-pr-1227.herokuapp.com/2017/01/06/preparing-for-the-trump-era-midwest-tour

Issue #1254, opened 2019-09-03
Refactor jquery out of non-admin JS


veganstraightedge:
There are only two pages (on the public facing site) that use Javascript:

- https://crimethinc.com/support
- live blog articles, of which there are only three:
  - https://crimethinc.com/live (J20)
  - https://crimethinc.com/airportblockades (airport blockades)
  - https://crimethinc.com/g20live (G20 in Hamburg)

That Javascript (now) only loads on those pages, instead of all pages and all articles.

Let's see if we can cut down the page load of those two pages (and a couple JS dependencies) by refactoring the JS to use browser native Javascript only, no jquery, no other libraries.

- [x] `support.js` https://github.com/crimethinc/website/blob/master/app/assets/javascripts/support.js
- [x] live blog article js (not all actually use jquery)
  - [x] https://github.com/crimethinc/website/blob/master/app/assets/javascripts/base.js
  - [x] https://github.com/crimethinc/website/blob/master/app/assets/javascripts/article_queue.js
  - [x] https://github.com/crimethinc/website/blob/master/app/assets/javascripts/articles.js
  - [x] https://github.com/crimethinc/website/blob/master/app/assets/javascripts/articles_poll.js

***

The [`admin.js`](https://github.com/crimethinc/website/blob/master/app/assets/javascripts/admin.js) site still uses jQuery (in Bootstrap) and that's just fine for now. Also, Bootstrap v5 is dropping jquery too. https://github.com/twbs/bootstrap/pull/23586

[SSfWD js](https://github.com/crimethinc/website/blob/master/app/assets/javascripts/steal-something-from-work-day.js) can also be ignored for now.

Andykmcc:
Hi, i'm happy to take care of this, this evening (pacific time).

veganstraightedge:
Thanks @Andykmcc! 

It also looks like @wbholland (thanks!) already PRed one of the files. 

Maybe you two can coordinate to divide and conquer the work?

wbholland:
Sorry, first time contributing to a repository and didn't think about claiming on here before I began work. I think only 2 other files (support.js and articles_poll.js) need refactoring. @Andykmcc feel free to grab these and even check over my PR if you'd like.

Andykmcc:
no worries @wbholland. i'll take a look at those other two files now and am happy to trade code reviews.

veganstraightedge:
@Andykmcc @wbholland Would y'all mind taking a look at this other JS file?

[...]

PR #1256, opened 2019-09-03
refactored to remove JQuery

wbholland:
Issue #1254

### DIFF
--- a/app/assets/javascripts/articles.js
+++ b/app/assets/javascripts/articles.js
@@ -1,13 +1,21 @@
 App.articleQueue.onIncoming = function(data) {
-  var alertDiv = $('#floating-alert');
+  var alertDiv = document.querySelectorAll('#floating-alert');
 
   if (alertDiv.length === 0) {
-    alertDiv = $('<div id="floating-alert" class="alert alert-floating alert-notice">)');
-    $('body').prepend(alertDiv);
+    alertDiv = document.createElement('div');
+    alertDiv.id = 'floating-alert';
+    alertDiv.className = 'alert alert-floating alert-notice';
+    
+    document.body.prepend(alertDiv);
   }
 
   var postPlural = data.length > 1 ? 'posts' : 'post';
-  var closeLink = $('<a class="close">&#10006;</a>').click(function() { alertDiv.remove() });
-
-  alertDiv.html('Psst! Refresh for ' + data.length + ' new ' + postPlural).append(closeLink);
+  
+  var closeLink = document.createElement('a');
+  closeLink.className = 'close';
+  closeLink.innerHTML = '&#10006;';
+  closeLink.addEventListener('click', function() { alertDiv.remove() });
+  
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/articles.js
```diff
@@ -1,13 +1,21 @@
 App.articleQueue.onIncoming = function(data) {
-  var alertDiv = $('#floating-alert');
+  var alertDiv = document.querySelectorAll('#floating-alert');
```
veganstraightedge: Since only one element on the page can have the id of `#floating-alert`, does it make more sense to use `getElementById()` instead of `querySelectorAll()`?

https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById

Andykmcc: since `#floating-alert` is an ID it will only ever be matched one element, so you might want to use `document.querySelector('#floating-alert')`. Then you can avoid some headache below.

app/assets/javascripts/articles.js
```diff
@@ -1,13 +1,21 @@
...
 
   if (alertDiv.length === 0) {
```
veganstraightedge: But if you change `querySelectorAll()` to `getElementById()`, you'd probably need a different conditional here.

app/assets/javascripts/articles.js
```diff
@@ -1,13 +1,21 @@
...
-    alertDiv = $('<div id="floating-alert" class="alert alert-floating alert-notice">)');
-    $('body').prepend(alertDiv);
+    alertDiv = document.createElement('div');
+    alertDiv.id = 'floating-alert';
+    alertDiv.className = 'alert alert-floating alert-notice';
```
Andykmcc: i normally use `alertDiv.setAttribute()` and `alertDiv.classList.add()`, etc for this sort of stuff. In this context i don't think it makes a difference so... 🤷‍♂. just my two cents.

app/assets/javascripts/articles.js
veganstraightedge: Or maybe just `querySelector()`

wbholland: True! I committed a change to querySelector() https://github.com/wbholland/website/commit/40b5707597c4db1d3930a620715bda37718d9b0b

app/assets/javascripts/articles.js
wbholland: Thanks! You're right, that's better style. I made a commit with the changes you suggested: https://github.com/wbholland/website/commit/40b5707597c4db1d3930a620715bda37718d9b0b

veganstraightedge:
Thanks @wbholland for the PR.
And @Andykmcc for the feedback on it.
🎉

Merging now.

veganstraightedge:
@wbholland  Feel free to open a PR to add yourself to the list of authors in the README (which is far from exhaustive).

PR #1257, opened 2019-09-04
Update articles.js

cilgintokatli:
updated some old jquery with vanilla js


app/assets/javascripts/articles.js
```diff
@@ -1,13 +1,20 @@
 App.articleQueue.onIncoming = function(data) {
-  var alertDiv = $('#floating-alert');
+  var alertDiv = document.getElementById('#floating-alert');
+  var body = document.querySelectorAll('body');
```
just1602: Is there a specific reason to use `querySelectorAll` instead of `querySelector` since there is only one `body` tag in the page?

app/assets/javascripts/articles.js
```diff
@@ -1,13 +1,20 @@
...
 
   if (alertDiv.length === 0) {
-    alertDiv = $('<div id="floating-alert" class="alert alert-floating alert-notice">)');
-    $('body').prepend(alertDiv);
+    alertDiv = document.querySelectorAll('<div id="floating-alert" class="alert alert-floating alert-notice">');
```
just1602: When I check the mdn documentation about `querySelectorAll` it says : 

> A DOMString containing one or more selectors to match against. This string must be a valid CSS selector string; if it's not, a SyntaxError exception is thrown.

So is this working to create an element? Shouldn't create the element using `alertDiv = document.createElement('div')` then `alertDiv.setAttribute`, etc ?

just1602:
Thanks for stepping in and taking this issue @cilgintokatli ! :tada:

I leave some questions, but it's nothing big. Let me know your thought on this and we'll be able to move forward with this. :)

app/assets/javascripts/articles.js
cilgintokatli: after another look i reckon `document.getElementsByTagName("body")[0]` would be the optimum way to select body.

app/assets/javascripts/articles.js
cilgintokatli: you are correct about querySelectorAll uh what was i thinking..

just1602:
Thanks for the changes @cilgintokatli !

@veganstraightedge it looks good to me !

app/assets/javascripts/articles.js
veganstraightedge: Will `document.body` get you what you need?

veganstraightedge:
@cilgintokatli Thank you for this PR! 

Unfortunately, another PR came in before yours and that's the one that got merged. Fortunately, that PR and this PR were very similar. 

https://github.com/crimethinc/website/pull/1256

I do like your use of a template string though.

https://github.com/crimethinc/website/pull/1257/files#diff-81c3015196e67c9da0003b5837b1853aR19

Would you mind updating this PR to just add that line to the new version of this file?

cilgintokatli:
@veganstraightedge no problem at all. i am glad that i've helped crimethinc. with solidarity.

veganstraightedge:
@cilgintokatli Would you mind rebasing and integrating your template string line into the new `articles.js`, please?

https://github.com/crimethinc/website/blob/master/app/assets/javascripts/articles.js#L19

cilgintokatli:
@veganstraightedge sorry i thought you asked me update the file to just add one line.

veganstraightedge:
@cilgintokatli I did. But I did a bad job communicating my intention. That one line is great, but it needs to be integrated into the existing file. I'm sorry for the miscommunication.

cilgintokatli:
@veganstraightedge no worries mate. i just opened a new pull request.

veganstraightedge:
Perfect 👍🏻

veganstraightedge:
@cilgintokatli  Feel free to open a PR to add yourself to the list of authors in the README (which is far from exhaustive).

PR #1258, opened 2019-09-04
issues-1254 - Remove jquery from support.js

Andykmcc:
https://github.com/crimethinc/website/issues/1254
    
Removes jQuery dependency from support.js. Sorry about the noisy diff, split view might be your best friend in this case.

### DIFF
--- a/app/assets/javascripts/support.js
+++ b/app/assets/javascripts/support.js
@@ -1,40 +1,57 @@
-$(function() {
-  // Maps 'amount_slider' _values_ to Stripe dollar _amounts_ ('quantity')
-  var mapping = {
-    "1":    1, "2":    2, "3":    3, "4":    4, "5":    5, "6":    6, "7":     7, "8":    8, "9":    9, "10":  10,
-    "11":  11, "12":  12, "13":  13, "14":  14, "15":  15, "16":  16, "17":   17, "18":  18, "19":  19, "20":  20,
-    "21":  21, "22":  22, "23":  23, "24":  24, "25":  25, "26":  30, "27":   35, "28":  40, "29":  45, "30":  50,
-    "31":  55, "32":  60, "33":  65, "34":  70, "35":  75, "36":  80, "37":   85, "38":  90, "39":  95, "40": 100,
-    "41": 105, "42": 110, "43": 115, "44": 120, "45": 125, "46": 130, "47":  135, "48": 140, "49": 145, "50": 150,
-    "51": 200, "52": 250, "53": 300, "54": 350, "55": 400, "56": 450, "57":  500, "58": 550, "59": 600, "60": 650,
-    "61": 700, "62": 750, "63": 800, "64": 850, "65": 900, "66": 950, "67": 1000
+;(function (){
+  // hand rolled version of $(document).ready
+  function ready(fn) {
+    if (document.readyState !== 'loading'){
+      fn();
+    } else {
+      document.addEventListener('DOMContentLoaded', fn);
+    }
   }
 
-  var slider = document.getElementById('amount_slider')
+  function readyCb() {
+    // Maps 'amount_slider' _values_ to Stripe dollar _amounts_ ('quantity')
+    var mapping = {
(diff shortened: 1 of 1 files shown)


Andykmcc:
FYI @wbholland

app/assets/javascripts/support.js
```diff
@@ -1,40 +1,57 @@
...
+  function ready(fn) {
+    if (document.readyState !== 'loading'){
+      fn();
+    } else {
+      document.addEventListener('DOMContentLoaded', fn);
+    }
   }
 
-  var slider = document.getElementById('amount_slider')
+  function readyCb() {
```
veganstraightedge: Thanks for this PR, @Andykmcc! 🎉

Quick question: what does `readyCb()` mean to you? It's not immediately obvious to me.

Andykmcc: This page was using the jQuery document on ready [shorthand](https://learn.jquery.com/using-jquery-core/document-ready/) (`$(function(){...})`). This replaces it. 

The script tags are loaded in the `<head>` of the html, this means they are also executed before the rest of the document is loaded. so we need to wait to trigger the code until everything is loaded, otherwise this script tries to run `document.querySelector('#js_stripe_form')` before the html element `js_stripe_form` is loaded. `readyCb` is the name i gave to the callback that gets called once the document fires the native "DOMContentLoaded" event. 

using some kind of DOM ready event (as you were) is a best practice can be made unnecessary for rails apps if the `<script>` tags that load the JS are moved to the end of the body, just before `</body>`. that way they are loaded and run only after all the HTML they try to interact with has already been loaded. basically, move [this](https://github.com/crimethinc/website/blob/master/app/views/shared/_head.html.erb#L9) to inside the shared footers [here](https://github.com/crimethinc/website/blob/master/app/views/layouts/application.html.erb#L26).

veganstraightedge: Gotcha. 👍🏻

Could we name that function without an abbreviation, please? So, there's no ambiguity by future developers.

Andykmcc: I renamed it and eliminated one level of misdirection in there. Hopefully, it is easier to read now.

astronaut-wannabe:
fyi, all, checking `hide whitespace` in the diff makes it easier to read
Screen Shot 2019-09-04 at 8 18 06 AM

app/assets/javascripts/support.js
```diff
@@ -1,40 +1,55 @@
...
-  // Maps 'amount_slider' _values_ to Stripe dollar _amounts_ ('quantity')
-  var mapping = {
-    "1":    1, "2":    2, "3":    3, "4":    4, "5":    5, "6":    6, "7":     7, "8":    8, "9":    9, "10":  10,
-    "11":  11, "12":  12, "13":  13, "14":  14, "15":  15, "16":  16, "17":   17, "18":  18, "19":  19, "20":  20,
-    "21":  21, "22":  22, "23":  23, "24":  24, "25":  25, "26":  30, "27":   35, "28":  40, "29":  45, "30":  50,
-    "31":  55, "32":  60, "33":  65, "34":  70, "35":  75, "36":  80, "37":   85, "38":  90, "39":  95, "40": 100,
-    "41": 105, "42": 110, "43": 115, "44": 120, "45": 125, "46": 130, "47":  135, "48": 140, "49": 145, "50": 150,
-    "51": 200, "52": 250, "53": 300, "54": 350, "55": 400, "56": 450, "57":  500, "58": 550, "59": 600, "60": 650,
-    "61": 700, "62": 750, "63": 800, "64": 850, "65": 900, "66": 950, "67": 1000
+;(function (){
```
astronaut-wannabe: 👍

veganstraightedge:
Thanks for the PR @Andykmcc! Merging now.

veganstraightedge:
Would you mind taking a look at this other JS file?

https://github.com/crimethinc/website/blob/master/app/assets/javascripts/articles_poll.js

veganstraightedge:
@Andykmcc Feel free to open a PR to add yourself to the list of authors in the README (which is far from exhaustive).

PR #1262, opened 2019-09-05
Remove jquery from articles_poll.js

wbholland:
per issue #1254


app/assets/javascripts/articles_poll.js
```diff
@@ -18,7 +18,7 @@ function handleUpdatedPost(collectionPost, data) {
   var i,
       posts = data.posts;
 
-  collectionPost.data('publishedAt', data.published_at);
+  collectionPost.setAttribute('data-publishedAt', data.published_at);
```
guitsaru: Data attributes shouldn't be camel-cased.
```suggestion
  collectionPost.dataset.publishedAt = data.published_at
```

veganstraightedge:
@wbholland Thanks for putting this together. That's the last of our non-admin JS that was using jQuery. 🎉

PR #1307, opened 2019-10-09
Finish Rails 6 upgrade

astronaut-wannabe:
I realized that we are not really on rails 6 until we update the
config to load rails 6 defaults

looking at the config flags in `new_framework_defaults_6_0.rb` I see
mostly email stuff and ActiveJob stuff, so i don't think we'd be
affected and it *might* be okay to rip off the band-aid all at once


config/application.rb
```diff
@@ -26,7 +26,12 @@ class Application < Rails::Application
     config.middleware.insert_after ActionDispatch::Static, Rack::Deflater
 
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.0
+    config.load_defaults 6.0
+
+    # by default `form_with` generates remote forms. We don't really
+    # use this right now, so it is easier to have it default to
+    # `false`
+    config.action_view.form_with_generates_remote_forms = false
```
astronaut-wannabe: documentation: https://guides.rubyonrails.org/configuring.html#configuring-action-view

PR #1350, opened 2019-11-11
Handle `publish` with no datetime

astronaut-wannabe:
Currently, if you set the `publication_status` to `'published'` for an
article, and then click `Save` **WITHOUT** setting a publication date
and time, you will get a `500` error.

We have decided to treat this scenario by setting the publication date
100 years in the future

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -122,7 +122,7 @@ module Admin
       )
     end
 
-    def handle_published_wo_datetime permitted_params
+    def handle_published_without_datetime permitted_params
       return if @article&.published?
 
       publish_in_100_years = permitted_params[:publication_status] == 'published' &&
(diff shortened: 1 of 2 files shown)


app/controllers/admin/articles_controller.rb
```diff
@@ -102,22 +102,36 @@ def article_params
       # to 'published'
       handle_publish_now_situation(permitted_params) if params[:publish_now].present?
 
+      handle_published_wo_datetime(permitted_params)
```
veganstraightedge:
```suggestion
      handle_published_without_datetime permitted_params
```

app/controllers/admin/articles_controller.rb
```diff
@@ -102,22 +102,36 @@ def article_params
...
       permitted_params.merge!(
         publication_status: 'published',
-        published_at:       Time.zone.now,
-        published_at_tz:    Time.zone.name
+        published_at:       time,
+        published_at_tz:    zone
       )
     end
+
+    def handle_published_wo_datetime permitted_params
```
veganstraightedge:
```suggestion
    def handle_published_without_datetime permitted_params
```

PR #1358, opened 2019-11-20
Baseline typography for 2020 theme

edbury:
**tl;dr**

- Removes font files we won't be serving
- Declares the full spectrum of available Alright Sans and Markazi weights/italics
- Establishes a simple vertical rhythm and modular scale
- Sets some baseline typography

**Unused Font Files**

I'm not strongly opinionated on this one - mainly tidying. We may have other reasons to store them here. As is, they would also be deployed to the web, which doesn't seem great.

**Font Declarations**

I did these as single families to keep things more manageable/expected. 

We may not need the full 100-900 weight spectrum down the road, but I figured it'd be nice to have the full gamut while this is in progress.

**Modular Scale**

I'm a big fan of relying on proportions and relative type instead of declaring each size manually - makes it easy to maintain vertical rhythm and change the entire scale at once.

I've also included a few spacing units we can use as we build things out. `$base-spacing` is equal to one line of body copy (`1rem * $base-line-height`). I added some `em` based ones as well, in case context-relative spacing is needed down the line.


edbury:
@veganstraightedge anything amiss here?

veganstraightedge:
@edbury Nope! It all looks great. I was just swamped at RubyConf this week.

I have a slight squirmishness to the link tag for the google fonts. It's fine for now. But let's have a conversation about trade offs sometime.

PR #1413, opened 2019-12-24
Migrate asset precompile to manifest.js

astronaut-wannabe:
Sprockets 4 and rails 6 defaults to using the manifest file to
compile assets

https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs

I moved the files from the old `assets.precompile` config to the new
style, but the `lite_mode` CSS was a weird one (it is and html file,
not a CSS file)

Need to make sure lite mode still works after this change (and maybe
make it a CSS file instead of html/erb??)

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,3 +1,16 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
 //= link_directory ../stylesheets .css
+
+//= link jquery.js
+//= link live_blog.js
+//= link support.js
+//= link tochangeeverything.css
+//= link admin.js
+//= link admin.css
+//= link steal-something-from-work-day.css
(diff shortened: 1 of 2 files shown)


veganstraightedge:
This looks great!

One question though. 
Is jquery.js needed anymore? I think this was an unremoved relic from the jqyery removal effort.

astronaut-wannabe:
@veganstraightedge hmm, lemme look into the jquery thing. I think you are correct

app/assets/config/manifest.js
```diff
@@ -1,3 +1,16 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
 //= link_directory ../stylesheets .css
+
+//= link jquery.js
```
astronaut-wannabe:
```suggestion
```

PR #1428, opened 2019-12-29
Try MSP-Greg/actions-ruby in place of actions/setup-ruby@v1

veganstraightedge:

`¯\_(ツ)_/¯ `

n/a

Replaces the first party `actions/setup-ruby@v1` action with a third party one, `MSP-Greg/actions-ruby@master`.

Just see if CI runs/passes.

This is because the GitHub owned action seems to be abandoned and doesn't have a Ruby version newer than `2.6.3`. We're running `2.6.5` in production, and will likely be on `2.7.0` soon too. And it really makes sense to run the same Ruby in CI as in production.

https://github.com/actions/setup-ruby/issues/42

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install Ruby
         uses: MSP-Greg/actions-ruby@master
         with:
-          ruby-version: '2.6.5'
+          ruby-version: '2.6.x'
 
       - name: Checkout
         uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


veganstraightedge:
Screen Shot 2019-12-29 at 1 13 21 AM

Still installed 2.6.3.

MSP-Greg:
@veganstraightedge

Sorry, I should make it clear that the action only installs current Windows MinGW Rubies.

There are some examples of using RVM in Actions, see the rvm job in:

https://github.com/ruby/openssl/blob/master/.github/workflows/workflow.yml

I may try to integrate RVM into actions-ruby for non Windows platforms.

PR #1431, opened 2019-12-29
Locale articles count

veganstraightedge:

https://media.giphy.com/media/SUFuaOfn7qXaE/giphy.gif

n/a

Adds a column to `Locale`: `articles_count`, defaulted to zero.
Populates it on the way out the door in the `languages controller`.
TODO later: update it in a callback when an article is saved.
It calculates `articles_count` on `@locales` in `languages controller` instead of calculating it in the view.
It rejects locales that have an `articles_count` of zero in `languages controller` instead of in the view.

View `/languages`. Does it work? Are there any locales with zero articles?

Nah.

### DIFF
--- a/app/controllers/languages_controller.rb
+++ b/app/controllers/languages_controller.rb
@@ -6,6 +6,12 @@ class LanguagesController < ApplicationController
     @body_id = 'languages'
     @locales = Locale.all
 
+    @locales.each do |locale|
+      locale.articles_count = Article.live.published.where(locale: locale.abbreviation).count
+    end
+
+    @locales = @locales.reject { |locale| locale.articles_count.zero? }
+
     render "#{Theme.name}/languages/index"
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/languages_controller.rb
```diff
@@ -6,6 +6,15 @@ def index
     @body_id = 'languages'
     @locales = Locale.all
 
+    @locales.each do |locale|
+      locale.articles_count = Article.live.published.where(locale: locale.abbreviation).count
```
astronaut-wannabe: does this populate and persist the column? or do you need to do something like `locale.update!(articles_count: count)`

veganstraightedge: It does not _populate and persist_. I was thinking we could do that in a separate PR as an `after_save` hook on `Article`, then remove this block altogether. 

We could also add a `.live` (or similar) scope on `Locale` to accomplish the same thing.

PR #1462, opened 2020-01-12
Add the link in header for locale specific articles

### LINKED ISSUE #1454: Add link to /languages/foo if on foo.crimethinc.com
If you're on a language/locale subdomain, we should provide a link to jump to that language collection page.

- `https://de.crimethinc.com` => `https://de.crimethinc.com/languages/deutsch`
- `https://es.crimethinc.com` => `https://es.crimethinc.com/languages/español`
- `https://fr.crimethinc.com` =>`https://fr.crimethinc.com/languages/français`
- `https://pt.crimethinc.com` => `https://pt.crimethinc.com/languages/português-brasileiro`

This implementation should not be hardcoded, so it continues to Just Work™ when we add new subdomains.

The link should be on the homepage and paginated views of the homepage. Not sure it should be on every page. I’m open to thoughts on this.

just1602:

resolves #1454 

It add a link in the header to the `languages/foo` articles where `foo` is the locale and `locale != 'en'`.

Test this on staging with a different languages and the link should show up.

### DIFF
--- a/app/views/2017/shared/_header.html.erb
+++ b/app/views/2017/shared/_header.html.erb
@@ -15,7 +15,14 @@
       <li class="nav-link"><%= link_to t("header.search"),  "/search" %></li>
       <li class="nav-link"><%= link_to t("header.about"),   "/about" %></li>
     </ul>
+    <% unless I18n.locale == I18n.default_locale %>
+      <ul>
+        <% @locale = LocaleService.find(locale: nil, lang_code: I18n.locale) %>
+        <li class="nav-link"><%= link_to t('header.locale_articles'), "languages/#{@locale.canonical}" %></li>
+      </ul>
+    <% end %>
   </nav>
+
 
   <%= link_to t("header.support"), [:support], class: "button" %>
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,6 +32,7 @@ en:
     support: Support Us
     search: Search
     about: About
+    locale_articles: "English's articles"
 
     first_time: First Time? Start Here →
     site_mode: Switch to the full version of the site →
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -32,6 +32,7 @@ fr:
     support: Support Us
     search: Search
     about: About
+    locale_articles: 'Articles en français'
 
     first_time: First Time? Start Here →
     site_mode: Switch to the full version of the site →


app/views/2017/shared/_header.html.erb
```diff
@@ -15,8 +15,15 @@
       <li class="nav-link"><%= link_to t("header.search"),  "/search" %></li>
       <li class="nav-link"><%= link_to t("header.about"),   "/about" %></li>
     </ul>
+    <% unless I18n.locale == I18n.default_locale %>
+      <ul>
+        <% @locale = LocaleService.find(locale: nil, lang_code: I18n.locale) %>
```
veganstraightedge: Can we move this to a `before_action` method in `ApplicationController`?

Maybe something like `set_current_locale`?

app/views/2017/shared/_header.html.erb
```diff
@@ -15,8 +15,15 @@
       <li class="nav-link"><%= link_to t("header.search"),  "/search" %></li>
       <li class="nav-link"><%= link_to t("header.about"),   "/about" %></li>
     </ul>
+    <% unless I18n.locale == I18n.default_locale %>
```
veganstraightedge: We could probably benefit from a blank line separating this thought and the previous group of nav links.

app/views/2017/shared/_header.html.erb
```diff
@@ -15,8 +15,15 @@
...
+      <ul>
+        <% @locale = LocaleService.find(locale: nil, lang_code: I18n.locale) %>
+        <li class="nav-link"><%= link_to t('header.locale_articles'), "languages/#{@locale.canonical}" %></li>
+      </ul>
+    <% end %>
   </nav>
 
+
```
veganstraightedge: Was this intentional? I'm not sure we need it.

config/locales/en.yml
```diff
@@ -32,6 +32,7 @@ en:
     support: Support Us
     search: Search
     about: About
+    locale_articles: "English's articles"
```
veganstraightedge: You can delete the wrapping `"` quotes. We don't need those in YAML (unless there's a `:` or something similar in the value).

config/locales/fr.yml
```diff
@@ -32,6 +32,7 @@ fr:
     support: Support Us
     search: Search
     about: About
+    locale_articles: 'Articles en français'
```
veganstraightedge: You can also delete the wrapping quotes here.

config/locales/en.yml
veganstraightedge: Can you use this text, please?

> See all articles in English →

veganstraightedge: We'll also need this exact same key in every YAML locale file. Even if we don't have a translation yet. Otherwise we get a "translation missing" error.

app/views/2017/shared/_header.html.erb
just1602: Not at all.

app/views/2017/shared/_header.html.erb
just1602: Fixed.

config/locales/en.yml
just1602: Good to know !

app/views/2017/shared/_header.html.erb
just1602: I'm not a big fan of callback like `before_action` and other since they have tendency to make testing harder and I think we should prefer extracting logic in service and / or query object instead of relying  on callback, but if you really want it to be in a callback, I'll refactor it. :)

PR #1464, opened 2020-01-12
Non english articles link improvments

just1602:

Improve previous merged PR #1454

It refactor some previously merged code to add link to specific locale articles.

Deploy to staging and test everything is still working with other locales.

Check #1462

# Acceptance Criteria

- [ ] The link should be still working as expected.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -11,6 +11,8 @@ class HomeController < ApplicationController
     # Feed artciles
     @articles = articles_for_current_page.page(params[:page]).per(6).padding(1)
 
+    @locale = LocaleService.find(locale: nil, lang_code: I18n.locale)
+
     render "#{Theme.name}/home/index"
   end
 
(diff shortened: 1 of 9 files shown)


config/locales/cz.yml
```diff
@@ -32,6 +32,7 @@ cz:
     support: Support Us
     search: Search
     about: About
+    locale_articles: See all articles in English →
```
just1602: I'll change this to Chinese I guess and other to the right locale.

app/controllers/home_controller.rb
```diff
@@ -11,6 +11,8 @@ def index
     # Feed artciles
     @articles = articles_for_current_page.page(params[:page]).per(6).padding(1)
 
+    @locale = LocaleService.find(locale: nil, lang_code: I18n.locale)
```
veganstraightedge: If we put this here, it'll only work on the homepage.

We need to put into `ApplicationController` and in a `before_action`, so it gets called on every request.

just1602: You're so damn right.. I was thinking about our discussion to only put it on the home page, but I put it in the header...

PR #1540, opened 2020-03-26
Start homepage content for 2020 theme

veganstraightedge:

alt text


app/controllers/home_controller.rb
```diff
@@ -8,8 +8,19 @@ def index
     # Homepage featured article
     @top_article = articles_for_current_page.first if first_page?
 
-    # Feed artciles
-    @articles = articles_for_current_page.page(params[:page]).per(6).padding(1)
+    if Theme.name == '2020'
```
just1602: It works well for now, but maybe we could think about a way to extract thise logic into a PORO instead of keeping it in the controller? Or creating ViewObject. Something like `HomePage` or `HomeFeed` or `HomePageFeed` ?

veganstraightedge: I'm open to that! After this PR lands, wanna take a swing at it?

just1602: I'm a little bit busy, but I'll follow what's going on and try to give some help !

PR #1545, opened 2020-04-01
Refactor Locale.live

veganstraightedge:


just1602:
Is there a reason why `Locale` contains a computed value instead of a relation to articles and we don't just use something like `Locale.all.includes(:articles)` when we need it for the locales page for example?

If we want to keep the computed value in the table, I'd suggest to avoid callback. I know in this example they allow us to DRY the code, but callbacks can lead to some spaghetti code. If we want to extract the logic from the controller we could a PORO that would do this job and call it during our actions or we can create a background job that'd do this job asynchronously. What are your takes about taht ?

astronaut-wannabe:
> Is there a reason why `Locale` contains a computed value instead of a relation to articles and we don't just use something like `Locale.all.includes(:articles)` when we need it for the locales page for example?

i assume for performance. that join gets costlier the more articles we have.

> If we want to keep the computed value in the table, I'd suggest to avoid callback. I know in this example they allow us to DRY the code, but callbacks can lead to some spaghetti code. If we want to extract the logic from the controller we could a PORO that would do this job and call it during our actions or we can create a background job that'd do this job asynchronously. What are your takes about taht ?

I am not too concerned about spaghetti code since our models are basically nothing but rails attributes and callbacks. not sure we really need to worry about async for the admin, but I am not opposed to that (running a job queue is more infrastructure to be maintained though)

I'd rather put something like this in a rails concern (like `Countable`) rather than a PORO. That is because this is a concern of the ActiveRecord model.

I usually reserve a poro for "business logic" that is less tightly coupled with the DB

all that said, this seems fine in `Articles` until we need to start counting other `Publishable` things

veganstraightedge:
Part of the problem is that at some point in the past, I decide to move away from a `locales` table and `locale_id` on `articles`. Instead, I thought it would be simpler / faster to just store the string of the locale's abbreviation / lang_code. `en` `de` etc.

I'm thinking now that it'd make sense to go back to Locales being a proper association to Articles. I think moving toward that (and moving some of the Locale/s Services into the database) will simply a lot of our localization effort.

PR #1549, opened 2020-04-04
Add Word .doc/docx to markdown upload and conversion tool

veganstraightedge:
To help publish translations that come in as Word files.


db/migrate/20200404041023_add_word_doc_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddWordDocToArticles < ActiveRecord::Migration[6.0]
+  def change
+    add_column :articles, :word_doc, :text
```
just1602, line 3: Do you need this at all if you delete the file after the conversion?

veganstraightedge: Nope!

We can rework this to not clear out the column before save. Or to not need the column at all.

Issue #1194, opened 2019-07-02
Add better Tag validation


veganstraightedge:
Right now, it's possible to create a tag that looks like a tag that already exists, but it gets save with leading/trailing blank space, or a different case… and ends up with a different slug.

For example:
```
  ["work", "Work"],
  ["work-1", "work"],
```

```
  ["anarchism", "  anarchism"],
  ["anarchism", " anarchism"],
  ["anarchism", "anarchism"],
```

The system should be smarter about what's new and what's an existing tag in a different shape.

```
[
  ["1917", "1917"],
  ["1917undead", "#1917undead"],
  ["2017", "2017"],
  ["2018", "2018"],
  ["911", "9/11"],
  ["accountability", "accountability"],
  ["activism", "Activism"],
  ["adventure", "adventure"],
  ["afd", "AfD"],
  ["alt-right", "alt right"],
  ["amazigh", "Amazigh"],
  ["analysis", "analysis"],
  ["anarchism", "  anarchism"],
  ["anarchism", " anarchism"],
  ["anarchism", "anarchism"],
  ["anarchisme", "anarchisme"],
  ["anarchist", "anarchist"],
  ["anarchist-arts", "anarchist arts"],
  ["anarchist-history", "anarchist history"],
  ["anarchy", "anarchy"],
  ["animal-enterprise-terrorism-act", "Animal Enterprise Terrorism Act"],
  ["anti-capitalism", "anti-capitalism"],
  ["anti-globalization", "anti-globalization"],
  ["anti-racism", "anti-racism"],
  ["antifa", "antifa"],
  ["antifascism", "antifascism"],
  ["apps", "apps"],
  ["argentina", "Argentina"],
  ["armed-struggle", "Armed Struggle"],
  ["armenia", "Armenia"],
  ["art", "art"],
  ["art-criticism", "art criticism"],
  ["athens", "Athens"],
  ["atlanta", "Atlanta"],
  ["auf-deutsch", "auf Deutsch"],
  ["authoritarianism", "Authoritarianism"],
  ["autonomous-spaces", "autonomous spaces"],
  ["baltimore", "Baltimore"],
  ["banner-drop", "banner drop"],
  ["bar", "bar"],
  ["barbacha", "Barbacha"],
  ["bay-area", "Bay Area"],
  ["belarus", "Belarus"],
  ["benefit", "benefit"],
  ["berkeley", "Berkeley"],
  ["betrayal", "betrayal"],
[...]

PR #1600, opened 2020-05-13
[#1194] Make tag name uniqueness case insensitive

just1602:

Didn't find the right gif.. :/

Make #1194 better

It makes the tag validation case-insensitive. This way, anarchism and Anarchism will be the same and it will avoid some tag duplication.

There is a unit test for it. Otherwise, you can create a resource with a tag named `Anarchism` and other one with the tag `anarchism` and the to resources will be link to the same tag.

# Acceptance Criteria

### DIFF
--- a/spec/models/tag_spec.rb
+++ b/spec/models/tag_spec.rb
@@ -3,6 +3,17 @@ require 'rails_helper'
 describe Tag do
   subject(:tag) { described_class.new(name: 'test') }
 
+  describe 'validation' do
+    context 'when two tag has the same name with case difference' do
+      it 'makes the second tag invalid' do
+        described_class.create!(name: 'anarchism')
+        second_tag = described_class.new(name: 'Anarchism')
+
+        expect(second_tag).to be_invalid
+      end
+    end
+  end
+
   describe 'assigned_to?' do
     let(:page) { Page.create(title: 'about') }
 
--- a/spec/models/tag_spec.rb
+++ b/spec/models/tag_spec.rb
@@ -4,7 +4,7 @@ describe Tag do
   subject(:tag) { described_class.new(name: 'test') }
 
   describe 'validation' do
-    context 'when two tag has the same name with case difference' do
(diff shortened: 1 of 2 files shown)


spec/models/tag_spec.rb
```diff
@@ -3,6 +3,8 @@
 describe Tag do
   subject(:tag) { described_class.new(name: 'test') }
 
+  it { should validate_uniqueness_of(:name).case_insensitive }
```
veganstraightedge, line 6: Is it possible to write this test without using the shoulda matchers?

just1602: It is. I really like this gem to write unit test for model validation and other model stuff like that, but if we don't plan to test this kind of stuff I can certainly remove shoulda-matchers and rewrite the test manually.

veganstraightedge: I guess I wonder if this is testing our code or testing Rails itself.

just1602: Fixed, if we want to test more models validation and relation maybe think at this gem, otherwise it sounds good I think.

spec/models/tag_spec.rb
```diff
@@ -3,6 +3,17 @@
 describe Tag do
   subject(:tag) { described_class.new(name: 'test') }
 
+  describe 'validation' do
+    context 'when two tag has the same name with case difference' do
+      it 'makes the second tag invalid' do
+        described_class.create!(name: 'anarchism')
+        second_tag = described_class.new(name: 'Anarchism')
+
+        expect(second_tag).to be_invalid
```
veganstraightedge, line 12: This is great! 🎉

spec/models/tag_spec.rb
just1602: Maybe we're not enough working on this codebase for this to make sense, but when you have a lot of people who works on a codebase, this is nice to avoid this kind of change to pass the code review by error. :)

PR #1618, opened 2020-05-24
[#1548] Use CurrentAttributes to replace current_user with Current.user

### LINKED ISSUE #1548: Use Current Attributes for: Current.user (admin)
This [also](https://github.com/crimethinc/website/issues/1547) feels like a place that we could benefit from using  `CurrentAttributes` (added in Rails 5.2).

They're _thread locale globals_. https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html  

`Current.user` would only apply to admin users, not readers.

veganstraightedge:
# #1548

***

🎉

resolves #1548

- Sets `Current.user` from the session
- Uses `Current.user` everywhere instead of `current_user`

Sign in and it should work.

Whenever we can use the framework itself and create/maintain less ourselves, the better.

### DIFF
--- a/app/views/admin/articles/_form.html.erb
+++ b/app/views/admin/articles/_form.html.erb
@@ -23,7 +23,7 @@
 
     <!-- TEMP: move to partial -->
     <%# TEMP: enable for alpha tolerant users, only on unpublished articles %>
-    <% if (1..4).include?(current_user.id) && @article.draft? %>
+    <% if Current.user.id == 1 && @article.draft? %>
       <fieldset class="card mb-3">
         <div class="card-body">
           <%= form.label :word_doc, "Upload a Word Doc file", class: "card-title m-0" %>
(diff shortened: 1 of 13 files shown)


app/views/admin/articles/_form.html.erb
```diff
@@ -23,7 +23,7 @@
 
     <!-- TEMP: move to partial -->
     <%# TEMP: enable for alpha tolerant users, only on unpublished articles %>
-    <% if (1..4).include?(current_user.id) && @article.draft? %>
+    <% if Current.user.id == 1 && @article.draft? %>
```
astronaut-wannabe, line 26: just checking, this is intentionally more restrictive?

veganstraightedge: Yeah. It's just me for now. Until the feature is actually working. 

Good catch. 👍🏻

PR #1623, opened 2020-05-25
[#1547] Add and use Current.locale

### LINKED ISSUE #1547: Use Current Attributes for: Current.locale
Now that we use the current person's locale all over the place, it'd be nice to standardize how we do it. And do it in a unified place

This feels like a place where we should use `CurrentAttributes` (added in Rails 5.2).

They're _thread locale globals_.
https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html

It would be swell if we could get/set `Current.locale` from anywhere in the app.

veganstraightedge:
# Closes #1547

Like https://github.com/crimethinc/website/pull/1621

Use `CurrentAttributes` for the current locale, which can be set from:

- subdomain
- user's browser setting
- or from the URL params (like in TCE).

### DIFF
--- a/app/middlewares/rack/current_locale.rb
+++ b/app/middlewares/rack/current_locale.rb
@@ -1,0 +1,86 @@
+module Rack
+  class CurrentLocale
+    def initialize app
+      @app = app
+    end
+
+    def call env
+      @req = Rack::Request.new(env)
+
+      set_current_locale
+      set_current_locale_from_subdomain
+      redirect_to_locale_subdomain
+
+      @app.call(env)
+    end
+
+    private
+
+    def set_current_locale
+      Current.locale = ::Locale.find_by abbreviation: I18n.locale
+    end
+
+    def set_current_locale_from_subdomain
+      return if subdomain.blank?
+      return unless I18n.available_locales.include? subdomain.to_sym
+
+      Current.locale = ::Locale.find_by abbreviation: subdomain
+    end
+
+    def redirect_to_locale_subdomain
+      # Force the subdomain to match the locale.
+      # Don’t do this in development, because typically local development
+      # environments don’t support subdomains (en.localhost doesn’t resolve).
+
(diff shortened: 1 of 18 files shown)


app/views/2017/shared/_header.html.erb
```diff
@@ -16,9 +16,9 @@
       <li class="nav-link"><%= link_to t("header.about"),   "/about" %></li>
     </ul>
 
-    <% unless Locale.english? %>
+    <% unless Current.locale.english? %>
       <ul>
-        <li class="nav-link"><%= link_to t('header.locale_articles'), language_path(@site_locale.canonical) %></li>
+        <li class="nav-link"><%= link_to t('header.locale_articles'), language_path(Current.locale.name) %></li>
```
astronaut-wannabe, line 21: I don't know if this is the source of the test failures, but `canonical` == `slug`

```suggestion
        <li class="nav-link"><%= link_to t('header.locale_articles'), language_path(Current.locale.slug) %></li>
```

app/middlewares/rack/current_locale.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      @app.call(env)
+    end
+
+    private
+
+    def set_current_locale
+      Current.locale = ::Locale.find_by abbreviation: I18n.locale
+    end
+
```
astronaut-wannabe, line 22: this is why the tests are failing.

You are making it so that the routing logic now depends on the locale database tables to exist an be populated

this will be `nil` in the specs because there are no locales, and that will make line 36 fail.

you can get the specs passing by putting `before { create(:locale, :en) }` at the top of the failing specs, but this brings up a question for me

should middleware routing be tied to the state of the DB? and does making a database call outside of application code introduce any risks (security or otherwise)

I am not familiar enough with rack to really know, but if it's fine then adding the `before` in failing specs should do the trick

PR #1631, opened 2020-05-31
[BUG] fix tag uniqueness logic 

astronaut-wannabe:

alt text

n/a

We have this validation on the `Tag` model:

```ruby
  validates :name, uniqueness: { case_sensitive: false }
```

this is because we want to treat a tag with a name of `"Foo"` the same
as a tag with the name of `"foo"`

however, we already have some tags that are upper-cased, some
lower-cased, etc.

We should do some sort of backfill to normalize all existing tags to
be downcased, but until then using an `lower(name)` postgres query
should make this work as intended

on the pr app, try drafting a new article with categories

https://crimethinc-staging-pr-1631.herokuapp.com/admin/articles

### DIFF
--- a/app/services/tag_assigner.rb
+++ b/app/services/tag_assigner.rb
@@ -2,8 +2,14 @@ class TagAssigner
   def self.parse_glob glob
     return if glob.blank?
 
-    names = glob.split(',').reject(&:blank?)
-    tags = names.map { |name| Tag.find_or_initialize_by(name: name.strip) }
+    names = glob.split(',').reject(&:blank?).map { |n| n.strip.downcase }.uniq
+    tags = names.map do |name|
+      # postgres treats everything as case-sensitive by default, so we
+      # cannot use find_or_initialize_by until we clean up the tags
+      # table and only have lowercase tag.names
+      existing_tag = Tag.where('lower(name) = ?', name).first
+      existing_tag || Tag.new(name: name)
+    end
     new(*tags)
   end
 
--- a/app/views/admin/articles/form/_categories.html.erb
+++ b/app/views/admin/articles/form/_categories.html.erb
@@ -5,7 +5,7 @@
       <p class="m-0 font-weight-bold">Categories</p>
 
       <div class="form-check">
-        <% @categories.each do |category| %>
+        <% @categories&.each do |category| %>
           <div>
             <%= form.label "category_ids_#{category.id}", class: "form-check-label" do %>
               <%= category_check_box form: form, category: category %>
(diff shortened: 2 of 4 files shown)


just1602:
@astronaut-wannabe I was sure the validation was done on the ruby side since postgres was using a case sensitive index and since the uniqueness validation can be use on column without index anyways. Am I wrong?

astronaut-wannabe:
> @astronaut-wannabe I was sure the validation was done on the ruby side since postgres was using a case sensitive index and since the uniqueness validation can be use on column without index anyways. Am I wrong?

The issue is that postgres is case sensitive, but our ruby logic and ruby ruby validation is assuming that it is *NOT* case sensitive

so we can have a tag with a name of `foo` in the database already, and if we try saving a tag called `Foo`, you will get a validation error that the name is not unique

that's all good so far, but if in the admin UI someone can enter `Foo`, and our code is expecting `find_or_initialize_by` to return the existing tag with name `foo`, but it does NOT do that (since those are unique names according to PG)

so it will initialize a new tag with `name: Foo`, and eventually try to `tag.save!`. causing the validation error (since ruby is validating `foo` and `Foo` case insensitively) 

we do not want a validation error. we want to find the `foo` tag when someone enters the `Foo` tag

astronaut-wannabe:
related to #1194

just1602:
Ok, so I thought I'd write a script, but after some investigation and debugging it'd be overkill!

All articles tagged with the weird named tags are also tagged with the right tag, so we can simply delete all the `Tagging` mapping article with the weirdly named `Tag` and then delete those tags and everything should be fine.

This should do the job:

```ruby
Tagging.where(tag_id: [2,3,4,5,6,7,8]).delete_all
Tag.where(id: [2,3,4,5,6,7,8]).delete_all
```

If you wanna know how I get those ids, here's what I did **on my machine with the latest data dump**:

```ruby
    grouped_tags = {}
    Tag.all.map { |t| grouped_tags[t.slug]=Array.new }
    # group tags by slug
    Tag.all.map { |t| grouped_tags[t.slug].push(t.id) }

    duplicated_tags = grouped_tags.filter { |k,v| v.length > 1 }
```

`duplicated_tags` should give you 4 slugs (anarchism, veganism, inside-front, hardcore) with different ids associated. The last id (the greater number) of each slug is the right one. So I delete all the `Tagging` and `Tag` of the other ids.

Once this is done it should be alright since the `Slug` concern validate the uniqueness of a slug before creating a new one.

If you want, I could check to get access and validate that it's the same data in produciton before doing the operation, but it should be.

just1602:
astronaut-wannabe's comment supression was an error!

Related [#1194]

PR #1638, opened 2020-06-04
Don't redirect en. subdomain anymore

veganstraightedge:
This is for people who have the browser / OS language set to something other than EN, but would like to see the site in English. They can't go to the apex domain because then the locale detection / subdomain redirect will kick in.

However, we don't want to redirect all EN apex requests to the `en.` subdomain, though.


veganstraightedge:
Thanks.

I'm gonna merge and test it on staging.
https://en.crimethincstaging.com

PR #1679, opened 2020-07-01
Sidekiq for Active Job

veganstraightedge:
Rails will use Sidekiq for background processing uploads.

https://guides.rubyonrails.org/active_storage_overview.html#analyzing-files


Procfile
```diff
@@ -1,2 +1,3 @@
-release: bundle exec rake db:migrate; echo $HDRIVE_GOOGLE_JSON_KEY > tmp/keyfile.json
+release: bundle exec rake db:migrate
 web: bundle exec puma -C config/puma.rb
+worker: RAILS_MAX_THREADS=${SIDEKIQ_RAILS_MAX_THREADS:-25} bundle exec sidekiq -q default -q mailers
```
bookis, line 3: I don't know this default value syntax, why the `-` in `-25`?

veganstraightedge: I hadn't seen it before, but it was in these docs:

https://devcenter.heroku.com/articles/concurrency-and-database-connections#background-workers

I guessed that it was a `Procfile` syntax to quack like 

`ENV.fetch('PREFERRED_SOURCE') { 'FALLBACK_VALUE' }

PR #1684, opened 2020-07-05
Skip rendering an article in the reader's preferred language when on languages#show

veganstraightedge:
[This PR](https://github.com/crimethinc/website/pull/1628) Just Worked™ in all places but one. That one is the exception. It should never be localized. Because it's on a language collection page.

https://github.com/crimethinc/website/pull/1628

https://github.com/crimethinc/website/commit/774fdddbf26f005e6f764115550ef4e406b5cb22#diff-8a12f2af3cd2315aff3cc6c928f6a46aR3

***

My solution was to explicitly skip it if that local is passed in.

```ruby
<% skip_preferred_localization ||= skip_preferred_localization %>
# ...
<%= render_themed "articles/list_item", 
                  article: skip_preferred_localization ? article : article.preferred_localization %>
```

### DIFF
--- a/app/views/2017/articles/_list.html.erb
+++ b/app/views/2017/articles/_list.html.erb
@@ -1,5 +1,7 @@
+<% skip_preferred_localization ||= skip_preferred_localization %>
+
 <div class="articles-list">
   <% articles.each do |article| %>
-    <%= render_themed "articles/list_item", article: article.preferred_localization %>
+    <%= render_themed "articles/list_item", article: skip_preferred_localization ? article : article.preferred_localization %>
   <% end %>
 </div>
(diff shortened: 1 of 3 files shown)


app/views/2017/articles/_list.html.erb
```diff
@@ -1,5 +1,7 @@
+<% skip_preferred_localization ||= skip_preferred_localization %>
```
astronaut-wannabe, line 1: is this an erb thing where you need to save to a variable to use it?

veganstraightedge: It's the pattern I've seen if you want to have a default. 

Ie, it's only set in this exceptional case. All other cases, no local is set or needed.

PR #1712, opened 2020-08-03
Add autofocus for the search field in search page

just1602:
Just a micro fix to make the search field in the search page autofocus when we hit the search page. :)


just1602:
Is there a reason why it's not there anymore? :P

https://github.com/crimethinc/website/blob/main/app/views/2017/shared/_site_search.html.erb#L3

veganstraightedge:
Oh hey. Yeah. We un-did it because it was also putting auto-focusing the search box on the homepage in the footer. 

So, when you went to homepage, it scrolled you to the bottom of the site. 

We should add this back but in a way that is /search page specific. Make sense?

just1602:
Oh totally, I didn't figure out it was doing this on the home page. :)

veganstraightedge:
No problem. We can pass in a local to the `render` of `autofocus: true/false` or something like that.

PR #1748, opened 2020-08-27
Add rake task to sync blobs

astronaut-wannabe:
adds a new rake task

### DIFF
--- a/lib/tasks/active_storage.rake
+++ b/lib/tasks/active_storage.rake
@@ -1,0 +1,41 @@
+namespace :active_storage do
+  desc 'Ensures all files are mirrored'
+  task mirror_all: [:environment] do
+    # All services in our rails configuration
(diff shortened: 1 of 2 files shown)


lib/tasks/active_storage.rake
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      # Upload local file to mirrors (if one exists)
+      if local_file.present?
+        mirrors.each do |mirror|
+          mirror.upload blob.key, local_file, checksum: blob.checksum
+        end
+      end
+
+      # If no local file exists then download a remote file and upload it to the mirrors (thanks @Rystraum)
```
astronaut-wannabe, line 31:
```suggestion
      # If no local file exists then download a remote file and upload it to the mirrors
```

PR #1756, opened 2020-09-03
Add pagination to tools

### LINKED ISSUE #1752: Add pagination to tools
image

Because the images are being loaded via Active Storage, they count as requests on the same domain. And because there are sooo many posters, some of the image requests are erroring with 429 TOO MANY REQUESTS.

https://httpstatuses.com/429

***

Changing the rate limit rack-attack rules could work, but I think this page has grown way too long and should be paginated anyhow. Stickers and Logos all being on one page still works because they're are not that many, but Posters (and Zines) needs it. So let's pick a per page threshold above `Sticker.count`.

just1602:
This PR aim at adding pagination to the tools pages, but first we have some refactor to do before we can paginate our tools.

resolves #1752 

This add pagination to tools pages

The pagination work when you visit the different tools page.

# Acceptance Criteria

The pagination is working on the tools pages

### DIFF
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -37,7 +37,7 @@ class ToolsController < ApplicationController
   end
 
   def set_tools
-    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published
+    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published.page(params[:page]).per(10)
   end
 
   def tool_class
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -6,7 +6,6 @@ class ToolsController < ApplicationController
   before_action :set_type
 
   before_action :set_tool,           except: :about
-  before_action :set_featured_tools, except: :about
   before_action :set_tools,          except: :about
 
   before_action :set_title
@@ -37,16 +36,8 @@ class ToolsController < ApplicationController
     @body_id = 'tools'
   end
 
-  def tools
-    tool_class.order(published_at: :desc).live.published
-  end
-
-  def set_featured_tools
-    @featured_tools = tools.english.where.not(buy_url: nil).map { |tool| tool if tool.buy_url.present? }.compact
-  end
-
   def set_tools
-    @tools = tools.english.map { |tool| tool if tool.buy_url.blank? }.compact
+    @tools = tool_class.order(position: :desc).live.published
   end
 
   def tool_class
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -37,7 +37,7 @@ class ToolsController < ApplicationController
   end
 
   def set_tools
-    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published
+    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published
   end
 
   def tool_class
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -37,7 +37,7 @@ class ToolsController < ApplicationController
   end
 
   def set_tools
-    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published.page(params[:page]).per(10)
+    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published.page(params[:page]).per(10)
   end
 
   def tool_class
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -37,7 +37,7 @@ class ToolsController < ApplicationController
   end
 
   def set_tools
-    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published.page(params[:page]).per(10)
+    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published.page(params[:page]).per(10)
   end
 
   def tool_class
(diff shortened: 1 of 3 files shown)


app/controllers/tools_controller.rb
```diff
@@ -37,16 +36,8 @@ def set_body_id
...
-    tool_class.order(published_at: :desc).live.published
-  end
-
-  def set_featured_tools
-    @featured_tools = tools.english.where.not(buy_url: nil).map { |tool| tool if tool.buy_url.present? }.compact
-  end
-
   def set_tools
-    @tools = tools.english.map { |tool| tool if tool.buy_url.blank? }.compact
+    @tools = tool_class.reorder(position: :desc).order(published_at: :desc).live.published
```
astronaut-wannabe, line 40:
```suggestion
    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published
```

this is need for the `preferred_localization` to work

app/controllers/tools_controller.rb
```diff
@@ -37,16 +36,8 @@ def set_body_id
...
-    tool_class.order(published_at: :desc).live.published
-  end
-
-  def set_featured_tools
-    @featured_tools = tools.english.where.not(buy_url: nil).map { |tool| tool if tool.buy_url.present? }.compact
-  end
-
   def set_tools
-    @tools = tools.english.map { |tool| tool if tool.buy_url.blank? }.compact
+    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published
```
astronaut-wannabe, line 40: just noticed this, we want `position` ascending (`published_at` is correct as `desc` I believe)

```suggestion
    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published
```

astronaut-wannabe (approved):
lgtm. i have #1766 ready, so maybe we dont need a backfill, we can just add the positions in admin after deploy

just1602:
AS you wish, the backfill is almost code and we'll need somethng because for now position is `nil` by default instead of being 0.

lib/tasks/data/update_position_on_tools.rake
```diff
@@ -0,0 +1,19 @@
+namespace :data do
+  desc 'Update position on tools'
+  task :update_position_on_tools => :environment do |_, args|
+    puts 'Updating zines position'
+    Zine.all.update_all(position: 10)
+    Zine.where(buy_url: nil).or(Zine.where(buy_url: "")).update_all(position: 0)
```
astronaut-wannabe, line 6: `@featured_tools` was all of the non-blank `buy_url`, right?

I think this will end up setting the previously featured tools to the bottom, but I think we want them at the top

I would just do 
```rb
Zine.where.not(buy_url: [nil, ""]).update_all(position: 1)
```

you could probably drop the `Zine.all.update_all(position: 10)` since if `position` is `nil` it will fall back to `published_at`

just1602: This is the actual feature vs normal tools, I'm not sure I understand correctly what the code is doing since it's fetching the nil buy_url, but it's checking if the url is present. https://github.com/crimethinc/website/blob/main/app/controllers/tools_controller.rb#L44-L50

astronaut-wannabe: the original code is saying 

```ruby
@featured_tools =  "all tools where buy_url is NOT nil or empty string"
@tools = "all tools where buy_url IS nil or empty string"
```
It doesn't really matter _why_ it was written with the odd usage of `.map` (maybe it was an effort to minimize DB calls?), what matters is that the end result is that tools with a buy_url show up at the top of the index page( e.g. `crimethinc.com/zines`).

the way this rake task is written, it will do the opposite. You set all zines to position 10, then you set zines **without** a `buy_url` to position 0.

0 comes before 10, we want the opposite. this only appears to work now because you changed the sortback to `desc` (i.e. 10 9 8 7 ..)

that is not how the position feature works on the rest of the site, it makes no sense for a user to put `100` in the admin UI, then have that show up at the top

app/controllers/tools_controller.rb
```diff
@@ -37,16 +36,8 @@ def set_body_id
...
-    tool_class.order(published_at: :desc).live.published
-  end
-
-  def set_featured_tools
-    @featured_tools = tools.english.where.not(buy_url: nil).map { |tool| tool if tool.buy_url.present? }.compact
-  end
-
   def set_tools
-    @tools = tools.english.map { |tool| tool if tool.buy_url.blank? }.compact
+    @tools = tool_class.english.reorder(position: :desc).order(published_at: :desc).live.published.page(params[:page]).per(10)
```
astronaut-wannabe, line 40:
```suggestion
    @tools = tool_class.english.reorder(position: :asc).order(published_at: :desc).live.published.page(params[:page]).per(10)
```

you changed this back to `desc`, it needs to be `asc`.

also, can you move it to the tools concern as a scope similar to https://github.com/crimethinc/website/blob/90111fa77150f4abd7f00f54a8b68bd41d1a3296/app/models/book.rb#L5

astronaut-wannabe (changes requested):

app/controllers/tools_controller.rb
just1602: Gosh, I'm sorry I really understand the position as the higher the position, the higher is the priority !

astronaut-wannabe: no worries :+1:

astronaut-wannabe (approved):
:ship:

just1602:
@astronaut-wannabe the linter error have already been fixed on master and I don't have the rights to merge, so if you want to merge this PR, go for it ! :)

astronaut-wannabe:
@just1602 ah, sorry! I thought the :heavy_check_mark: would unlock the button

PR #1762, opened 2020-09-08
Update cs.yml

noticedneko:
First version of infrastructure translation to czech language . I will send new pull request if i find any mistake

# How does this pull request make you feel (in animated GIF format)?

alt text

# What are the relevant GitHub issues?

resolves `[link to github issue]`

# What does this pull request do?

??

# How should this be manually tested?

??

# Is there any background context you want to provide for reviewers?

??

# Questions for the pull request author (delete any that don't apply):
- [ ] Are any needed README/wiki/documentation updates needed for this pull request?
- [ ] Does the code you updated have tests? If not, could you add some please?
- [ ] Does this pull request require any new server dependencies which need to be added to the build process? If so, please explain what.

# Acceptance Criteria
## These should be checked by the reviewers

- [ ] This pull request does not cause the database export script to become out of sync with the db schema

PR #1763, opened 2020-09-08
Update cz.yml

noticedneko:
first version of czech translation

alt text

resolves `[link to github issue]`

??

??

??

# Acceptance Criteria

### DIFF
--- a/config/locales/cz.yml
+++ b/config/locales/cz.yml
@@ -8,7 +8,7 @@
 #  8. If a value has Markdown syntax characters in it, please keep those in your translation.
 #  9. Markdown examples: [link text](url), **bold**, _italics_, %{dynamic variable}, etc.
 # 10. If a value has <b>HTML</b> syntax characters in it, please keep those in your translation.
-en:
+cz:
   name: Czech
   language_direction: ltr
   site_name: CrimethInc.
--- a/config/locales/cz.yml
+++ b/config/locales/cz.yml
@@ -376,6 +376,7 @@ cz:
         heading: Díky za pomoc!
         description: |
           Uvítáme tvou [zpětnou vazbu](mailto:info@crimethinc.com) tykající se naší snahy a tvé [účasti](mailto:info@crimethinc.com) v ní.
+
           Brzy ti přijde potvrzovací email.
 
           Jestli si se přihlásil k měsíčnímu předplatnému, můžeš ho upravit nebo zrušit kdykoliv chceš vyplněním tvé emailové adresy ve formuláři na spodu stránky [Podpora](/support) page.
--- a/config/locales/cz.yml
+++ b/config/locales/cz.yml
@@ -8,7 +8,7 @@
 #  8. If a value has Markdown syntax characters in it, please keep those in your translation.
 #  9. Markdown examples: [link text](url), **bold**, _italics_, %{dynamic variable}, etc.
 # 10. If a value has <b>HTML</b> syntax characters in it, please keep those in your translation.
-cz:
+en:
   name: Czech
   language_direction: ltr
   site_name: CrimethInc.
@@ -16,30 +16,30 @@ cz:
 
   head:
     # meta_title is wrapped in quotes because it has a colon in its text value
-    meta_title: "CrimethInc. ex-Workers’ Collective: Your ticket to a world free of charge"
+    meta_title: "CrimethInc. ex-Workers’ Collective: Tvůj lístek ke změně světa"
     meta_image_url: https://cloudfront.crimethinc.com/assets/share/crimethinc-site-share.png
-    meta_description: CrimethInc. is a decentralized network pledged to anonymous collective action. We strive to reinvent our lives and our world according to the principles of self-determination and mutual aid.
+    meta_description: CrimethInc. je decentralizovaná síť která se zavázala k anonymní kolektivní akci. Usilujeme o znovuobjevení našich životů a světa dle principů sebeurčení a vzájemné pomoci.
 
   header:
-    languages: Languages
-    books: Books
-    library: Library
-    tools: Tools
-    audio: Podcasts
+    languages: Jazyky
+    books: Knihy
+    library: Knihovna
+    tools: Nástroje
+    audio: Podcasty
     video: Video
-    store: Store
-    subscribe: Subscribe
-    support: Support Us
-    search: Search
-    about: About
-    locale_articles: See all articles in Czech →
+    store: Obchod
+    subscribe: Odebírej
+    support: Podpoř nás
+    search: Hledej
+    about: O nás 
+    locale_articles: Všechny články v českém jazyce →
 
-    first_time: First Time? Start Here →
-    site_mode: Switch to the full version of the site →
+    first_time: Jsi tu poprvé? Začni tady →
+    site_mode: Přepni pro plnou verzi stránky →
 
   search:
-    form_placeholder: Search Archives…
-    search_results: 'Search results for: %{query}'
+    form_placeholder: Prohledej Archiv…
+    search_results: 'Výsledky vyhledávání pro: %{query}'
     advanced:
       heading: Advanced Search
       form:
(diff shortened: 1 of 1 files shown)


config/locales/cz.yml
```diff
@@ -8,38 +8,38 @@
 #  8. If a value has Markdown syntax characters in it, please keep those in your translation.
 #  9. Markdown examples: [link text](url), **bold**, _italics_, %{dynamic variable}, etc.
 # 10. If a value has <b>HTML</b> syntax characters in it, please keep those in your translation.
-cz:
+en:
```
astronaut-wannabe, line 11: not sure if this is copypasta, but i think it still needs to be `cz`
```suggestion
cz:
```

config/locales/cz.yml
```diff
@@ -80,106 +80,107 @@ cz:
...
+        description:  Sleduj naše aktivity na následujících platformách.
 
     mantra:
-      - If your heart is free, the ground you stand on is liberated territory. Defend it.
-      - Only those prepared to go too far will learn how far they can go.
-      - Desperate measures are long overdue.
-      - Cradle the seed, even in the volcano’s mouth.
-      - The future is in chains within you — set it free.
-      - A spark for every powder keg.
+      - Jestli je tvé srdce volné, zem na které stojíš je již osvobozená. Nenech si ji vzít.
```
astronaut-wannabe, line 137: I don't read Czech, so this is more just me being curious. Feel free to ignore this comment:

Is this an idiom in Czech, just wondering as the dictionary I am using (https://slovnik.seznam.cz/preklad/cesky_anglicky/Nenech) kind of translated this one odd

config/locales/cz.yml
```diff
@@ -80,106 +80,107 @@ cz:
...
+      - Jen ti připraveni dojít příliš daleko zjistí jak daleko mohou jít 
+      - Zoufalé časy už dávno skončily
+      - Starej se o semínko, i ve chřtánu sopky.
+      - Budoucnost uvnitř tebe je v řetězech — osvoboď ji.
+      - Jiskra pro tvuj střelný prach.
       - Esperanza no espera.
       - Mobilis in mobili
+      - Bojujeme další válku.
```
astronaut-wannabe, line 145: is this one a czech saying? Just wondering

config/locales/cz.yml
```diff
@@ -225,184 +226,180 @@ cz:
...
         support_request:
-          header: Existing Monthly Subscribers
-          description: Need to update or cancel your subscription? Enter the email address you used to sign up, and we will send you an email with a super secret link where you can make changes.
-          email_label: Email Address
-          email_placeholder: you@example.com
-          button_text: Send Super Secret Link
+          header: Existující měsíční předplatné.
+          description: Potřebuješ upravit nebo zrušit předplatné? Vlož email kterým se přihlašuješ, A my ti na něj pošleme supertajný odkaz kde ho můžeš upravit .
+          email_label: Emailová adresa
+          email_placeholder: ty@příklad.com
```
astronaut-wannabe, line 339: oh this is interesting, `example.com` is actually a [set-aside domain for documentation][0], not sure if we should translate it. @veganstraightedge, thoughts?

[0]:https://en.wikipedia.org/wiki/Example.com

config/locales/cz.yml
```diff
@@ -225,184 +226,180 @@ cz:
...
+        notice: Tvé předplatné bylo zrušeno.
+        error: Při rušení tvého předplatného se něco nepovedlo. Zkus to znovu! Jestli problém přetrvává, prosím [pošli nám email na](mailto:info@crimethinc.com) abychom ti mohli pomoct.
 
       thanks:
-        heading: Thanks for your help!
+        heading: Díky za pomoc!
         description: |
-          We welcome your [feedback](mailto:info@crimethinc.com) about our efforts and your [participation](mailto:info@crimethinc.com) in them.
+          Uvítáme tvou [zpětnou vazbu](mailto:info@crimethinc.com) tykající se naší snahy a tvé [účasti](mailto:info@crimethinc.com) v ní.
+          Brzy ti přijde potvrzovací email.
```
astronaut-wannabe, line 379:
```suggestion

          Brzy ti přijde potvrzovací email.
```

config/locales/cz.yml
```diff
@@ -225,184 +226,180 @@ cz:
...
+          repeat_request: Jestli nestihneš co si potřeboval, není se čeho bát, vždy ti můžeme poslad další odkaz
+          more_help: Můžeš taky odpovědět na tento email jestli potřebuješ s něčím pomoct.
 
     languages:
-      view_site_in_locale:  View this site in Czech [here](https://cs.crimethinc.com).
-      view_tools_in_locale: You can access an array of posters, zines, and other materials in Czech [here](https://cs.crimethinc.com/tools).
-      view_books_in_locale: You can access our books in Czech [here](https://cs.crimethinc.com/books).
+      view_site_in_locale:  Podívej se na tuto stránku v angličtině [zde](https://crimethinc.com).
+      view_tools_in_locale: Tady najdeš řadu plakátů, zinů, a dalších materiálů v Angličtině  [zde](https://crimethinc.com/tools).
+      view_books_in_locale: Přístup k našim knihám v angličtině [zde](https://crimethinc.com/books).
```
astronaut-wannabe, line 394: I think these still need to point at the `cs` subdomain. @veganstraightedge do you know for sure?

config/locales/cz.yml
```diff
@@ -225,184 +226,180 @@ cz:
...
-        primer, [To Change Everything](/tce), is available in 30. We have
-        [long](/2010/08/18/non-english-crimethinc-projects) aspired to collect all the different
-        translations on this site. If you would like to translate our material—or if you can
-        help us to collect translations for this archive or help us maintain communication
-        between related projects around the world—please
-        [contact us](mailto:foreignlegion@crimethinc.com).
-
-        Here, you can find all the articles on this site categorized according to language.
+        CrimethInc. materiály se objevily ve více než třech desítkách jazyků; náš úvodní text, [Změňit vše](/tce), je dostupný ve třiceti.
+        [Dlouho](/2010/08/18/non-english-crimethinc-projects) se snažíme přeložit tuto stránku. Jestli by si chtěl přeložit náš materiál—nebo jestli nám můžeš pomoct najít překlady, nebo nám můžes pomoct v udržení komunikace s příbuznými projekty na celém.
```
astronaut-wannabe, line 399: The "we have" in the english version seems to have been dropped here between "třiceti" and "Dlouho"

astronaut-wannabe (commented):
@noticedneko thanks for doing this! very exciting

config/locales/cz.yml
noticedneko: it's not idiom, just freely translated from this "If your heart is free, the ground you stand on is liberated territory. **Defend it"** Only difference is that i changed the bold part from defend it which would be something like "Braň ho" and just sounded wrong in this context so i choose more freeflow tranlastion which didn't change the meaning , i don't thing i can translate it back without it sounding dumb but it would be something like "Don't let them take it from you"

config/locales/cz.yml
noticedneko: @astronaut-wannabe  "We have" isn't necessary here  the "we" part is included in "snažíme" which could be translated into english like "we are trying"  If that's what you're refering to

config/locales/cz.yml
noticedneko: Nah, nothing that interesting i'am afraid :/ , It's just loosely translated to czech language.

config/locales/cz.yml
astronaut-wannabe: sounds good to me!

veganstraightedge:
Thank you for doing doing both of these translation PRs. But, as it turns out, the CZ one isn't used. CZ redirects to CS. 

We have CZ historically because of a wrong assumption that I made when we were first adding non-EN languages to the site. We now use CS for all things Czech. I'll close this PR and we can focus on the other one instead:

# #1762

PR #1779, opened 2020-09-20
Less loops for related articles

astronaut-wannabe:

alt text

n/a

refactors the `Article#related` method to remove the nested loop

see if related articles still look right

no

- [X] Are any needed README/wiki/documentation updates needed for this pull request?

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -119,7 +119,7 @@ class Article < ApplicationRecord
     articles_to_exclude = [id]
     categories.each_with_object({}) do |category, hash|
       # get 3 articles that we haven't seen yet in a previous iteration of this loop
-      related_articles = category.articles.english.published.limit(3).where.not(id: articles_to_exclude)
+      related_articles = category.articles.english.live.published.limit(3).where.not(id: articles_to_exclude)
 
       # save the IDs of these articles so we can exclude them in the next iteration of this loop
       articles_to_exclude += related_articles.pluck(:id)
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -73,16 +73,17 @@ class Article < ApplicationRecord
   def related
     return {} if categories.blank?
 
-    related_articles = {}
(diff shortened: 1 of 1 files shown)


app/models/article.rb
```diff
@@ -71,24 +71,7 @@ def content_rendered include_media: true
...
-
-          articles << article
-        end
-
-        related_articles[category] = articles
-      end
-    end
-
-    related_articles
+    @related ||= find_related_articles
```
just1602, line 74: Does rails keep the models in memory between request? What I mean here is, do we call `related` twice on the same object in a single request?

I'm not against this change, its more a knowledge question.

app/models/article.rb
```diff
@@ -129,6 +112,22 @@ def lede
...
   private
 
+  def find_related_articles
+    return {} if categories.blank?
+
+    # we do not want the current article to show up in related articles
+    articles_to_exclude = [id]
+    categories.each_with_object({}) do |category, hash|
+      # get 3 articles that we haven't seen yet in a previous iteration of this loop
+      related_articles = category.articles.english.published.limit(3).where.not(id: articles_to_exclude)
```
just1602, line 122: In a lot of place in the codebase we use `live.published` is there a reason why here we only use `published`?

astronaut-wannabe: no, idea. I just copied the existing code. I think you are correct, let's throw `live` in there 

```suggestion
      related_articles = category.articles.english.live.published.limit(3).where.not(id: articles_to_exclude)
```

app/models/article.rb
astronaut-wannabe: Activerecord does do query caching, and we also do view caching, so this will only really affect un-cached articles

We do call `#related` more than once per request in the view

just1602 (approved):

veganstraightedge (approved):
Very pretty! A nice improvement over what we had before. 🎉

PR #1781, opened 2020-09-20
[REFACTOR][I18N] Use `translatable` concern with articles

astronaut-wannabe:

alt text

n/a

Now that the `Translatable` concern has been in use in production on
`/tools` pages for awhile, it seems safe to move the `Article`
implementation to also use the concern

check article translation workflows on staging

### DIFF
--- a/app/models/concerns/translatable.rb
+++ b/app/models/concerns/translatable.rb
@@ -24,22 +24,21 @@ module Translatable
 
   def localizations
     all_localizations = [
-      canonical_tool,
-      canonical_tool_localizations,
+      canonical_record,
+      canonical_record_localizations,
       self_localizations
     ]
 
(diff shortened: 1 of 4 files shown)


app/models/concerns/translatable.rb
```diff
@@ -24,22 +24,21 @@ def localization_in locale
...
 
-  def canonical_tool
+  def canonical_record
     self.class.find_by(id: canonical_id)
   end
 
-  def canonical_tool_localizations
-    canonical_tool&.localizations
+  def canonical_record_localizations
+    canonical_record&.localizations
```
astronaut-wannabe, line 41: just removing the `tool` specific naming and replacing it with a more generic `record` name

just1602 (approved):
This is really clean ! :)

PR #1867, opened 2020-11-12
Remove JQuery from the tweet length counter

### LINKED ISSUE #1863: Convert jQuery to vanilla JS in admin character counter (for admin form text areas)
If we convert this to vanilla Javascript, we'll be one step closer to removing jQuery as an admin dependency. 

https://github.com/crimethinc/website/blob/e0f5f241cb78d3f47833683cd3ce7a0e5f1d2ab0/app/assets/javascripts/admin.js#L9-L23

astronaut-wannabe:

alt text

resolves #1863 

removes the JQuery needed to update the tweet and description character counters

I also updated our modernizr library 

see the counters still work in admin

- [X] Does the code you updated have tests? If not, could you add some please?


app/assets/javascripts/admin.js
```diff
@@ -5,24 +5,29 @@
 //= require jquery_ujs
 //= require popper
 //= require bootstrap
+//= require vanilla_dom_ready
```
veganstraightedge, line 8: Could we name this simply `dom_ready` since we'll end up in a place where all of our JS is _vanilla_?

veganstraightedge (approved):

PR #1903, opened 2020-12-11
Remove jQuery, use Rails UJS in /admin

veganstraightedge:

So long and thanks for all the memories

- Removes `jQuery` from the last manifest that was still using it: `admin.js`
- Removes `jquery_ujs` which used `jquery` and handled the "Are you sure?" confirmations when deleting a record from `/admin`
- Adds `rails-ujs` to do what`jquery_ujs` did, but without `jquery` https://github.com/rails/rails/tree/master/actionview/app/assets/javascripts

- Go to `/admin`
- Edit any record (eg, article)
- Scroll down to The Danger Zone™
- Click the `Delete` button
- **Expected**:  confirmation dialog should popup and prevent deleting unless you click `OK`

image

We're stepping closer to no jQuery on the site and toward Bootstrap 5 which no longer depends on jQuery. 

# Acceptance Criteria

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -1,8 +1,7 @@
-//= require jquery3
-//= require jquery_ujs
 //= require popper
 //= require bootstrap
 //= require dom_ready
+//= require rails-ujs
 
 // character counter on textareas
 function updateCounter(e) {


just1602:
This is neat that the ujs thing is just an adapter an you can choose the implementation you want!

just1602 (approved):

astronaut-wannabe (approved):
lgtm!

| demo app |
|------|
| Screenshot from 2020-12-12 21-36-45 |

The docs mention a [yarn install][0], is that not needed for reasons?

[0]:https://github.com/rails/rails/tree/master/actionview/app/assets/javascripts#installation

veganstraightedge:
In our case, we don't have to because we're not using webpack/er for our JS assets. I think!

> **Asset pipeline**
> 
> In a conventional Rails application that uses the asset pipeline, require rails-ujs in your application.js manifest:
> 
> `//= require rails-ujs`

Issue #1912, opened 2020-12-20
[TCE] Replace jQuery with vanilla JS


veganstraightedge:
_To change everything… start everywhere_ (that there's deletable code)!

In the spirit of [removing more Javascript](https://github.com/crimethinc/website/issues/1254) from the site and reducing the page load where we can…

***

https://github.com/crimethinc/website/blob/1078fb874181cf77eeead05c4c84435ff48ab677/app/views/layouts/to_change_everything.html.erb#L12-L20

There are a couple functions that use jQuery but could be replaced with vanilla JS. Let's do that.

This is toward the goal of removing jQuery from TCE entirely. But the scope of this issue is **only** these couple functions.

PR #1920, opened 2020-12-24
Replace jQuery with vanilla JS [#1912]

QueerSynd:
# For #1912 

As discussed [here](https://github.com/crimethinc/website/commit/ce3cccda5e72e11473ddd83eca4d657a2240d7fd), I have adapted the resize callbacks to be run up to IE8.

**Suggestion:**
I would suggest to remove the resize events altogether, they shouldn't be of any use on the page. I investigated a bit and it feels like legacy code. The Z-Index applies to titles (there are no blockquotes on the page) that should already be positioned correctly (and it also applies to an element in the header which shouldn't be concerned by this script in my opinion. Removing the functions doesn't seem to affect the layout _as far as I've seen._

### DIFF
--- a/app/views/layouts/to_change_everything.html.erb
+++ b/app/views/layouts/to_change_everything.html.erb
@@ -9,20 +9,21 @@
   <%= stylesheet_link_tag "to_change_everything", media: "screen" %>
 
   <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
-  <script>    // For IE < 9
+  <script>
+    var resizeCallback = function () {
+      var elements = document.querySelectorAll("h1,h2,h4,blockquote")
+      for (var i = 0; i < elements.length; i++) {
+        elements[i].style.zIndex = '1'
+      }
+    }
+
+    // For IE8
     if(window.attachEvent) {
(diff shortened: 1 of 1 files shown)


astronaut-wannabe (approved):
lgtm

> I investigated a bit and it feels like legacy code. The Z-Index applies to titles (there are no blockquotes on the page) that should already be positioned correctly (and it also applies to an element in the header which shouldn't be concerned by this script in my opinion. Removing the functions doesn't seem to affect the layout as far as I've seen.

You are right about it being legacy. we have slowly been transitioning the TCE pages from static html files, into railsy erb templates

it's possible that the callbacks were doing something that is now entirely handled in CSS (the css is also in a state of transition, so there may be a lot of dead css code too)

veganstraightedge (approved):
@QueerSynd Thank you for this PR! 🎉

I echo what @astronaut-wannabe said. There is definitely legacy code handing around in TCE still as we slowly transition it forward. 

These PRs at least get us closer to one goal: being able to drop jQuery on TCE!

In a separate PR, let's explore doing what you suggest and dropping this resize listener altogether. 👍🏻

In the meantime, I'll merge and deploy this.

Issue #1924, opened 2020-12-26
[TCE] Un-modal video


veganstraightedge:

QueerSynd:
What about the current "watch video" cta, should it be removed and the download PDF centered?

astronaut-wannabe:
@QueerSynd good question. I think the PDF cta would look weird centered by itself.

the actual goal is to drop jquery, so in theory we could keep the current style and do it all in vanilla js (honestly, we could probably do it all in css)

all that said, i do think the new layout @veganstraightedge proposed for the video would be nice (because it would be easier to hide for languages where we don't yet have a translation)

@veganstraightedge what do you think about the solo pdf cta? We actually have this in [czech][0] (css is janky there since it is still being served out of /public, but that will be fixed in #1904)

also, this just came to me while thinking through this, if we add a new row for the video, we could have a new CTA in that row requesting translations of the script (maybe link to a GH issue with the english script broken out line by line?)

[0]:https://crimethinc.com/tce/czech

veganstraightedge:
A solo "download PDF" is fine for now. After it's solo, we can consider it if it feels weird by itself. And if so, we can reimagine how to handle it. 👍🏻

QueerSynd:
Hi, I've been out of the loop for a while, what is the status of this issue? Anything I can do to help close it?

PR #1931, opened 2020-12-31
[TCE] Un-modal video #1924

QueerSynd:
From [#1924](https://github.com/crimethinc/website/issues/1924)

**Outcome**
- Removed fancybox and JQuery from the page. 
- Removed the "Watch Video" button and it's translations (as discussed, the "Download PDF" button is left alone, centered).
- Moved the video in an iframe right in a new section below the titles.

| Width        | Screenshot           |
| ------------- |:-------------:|
|   1200px+   |   |
| 914px     |     |
| 686px      |  |
| 500px |    |

We should discuss the size and positioning of the video on various breakpoints, I just did it as a proposition as I am not sure what you might actually have in mind for smaller devices for example 😄 

**PS**: Sorry for the commits from other issues tagging along, I wasn't expecting them to show up after merging and pulling on my side. Will avoid the issue next time


veganstraightedge:
@QueerSynd Great start! Thanks for digging into this. Here are my thoughts about how to handle the different breakpoints:

**Video max width**, even at a larger window width: **1285px**

**Spacing around the video** in the black row: **at least 30px**, but maybe up to 50-100px vertically. We'll need to feel it out.

For **Download PDF**, let's try either:

1. Dropping the round rect and instead jump straight to the full width translucent white bar
2. Make the round rect only as big as the area of _Download PDF_, not `display: block` / full width

***

Does that all make sense?

Let's give those a try and see how it looks behaves then.

veganstraightedge:
@QueerSynd If you'd like to be added to our Slack, send an email to tech at our domain dot com with the email address that you want to be invited at.

QueerSynd:
Thanks for the feedback @astronaut-wannabe !

I just pushed the first solution **Download PDF** background, I think it is the most elegant of both. 

On the breakpoints though I am not sure I get it well:
> **Video max width**, even at a larger window width: **1285px**
> 
> **Spacing around the video** in the black row: **at least 30px**, but maybe up to 50-100px vertically. We'll need to feel it out.

I can set a max width, but the video being in an iframe, the size of the video is actually controlled by its height. What I can do is give the iframe a margin of 30px (even on smaller breakpoints? the video might be very small at this point), and control the height. I would propose something like that:

```css
max-width: 1285px;
height: 40vh;

@media (min-width: 576px) {
  height: 50vh;
}

@media (min-width: 900px) {
  height: 70vh;
}

@media (min-width: 1200px) {
  height: 95vh;
}
```
The place of the margin is left to be defined on my side.

what do you think?

QueerSynd:
Zsdert67uhbn

QueerSynd:
I committed the code as I suggested above.
I think this can be merged if there are no bugs, and we could open another ticket to define the size of the video precisely. What do you think?

veganstraightedge:
Ok, @QueerSynd. I merged the CSS listing PR. Would you mind rebasing this one, please? And then we can get it merged in too. 🎉

QueerSynd:
As you can see with the tones of commits, I've rebased @veganstraightedge 😅 I double checked quickly, the TCE page still looks as expected, but maybe you want to have a quick look too before merging.

veganstraightedge:
@QueerSynd Did you merge `main` into your branch or did you rebase your branch off of `main`? It's a subtle difference.

QueerSynd:
I rebased my branch, and github was still having conflicts for whatever reason, so I solved them (which is the last commit)

PR #1944, opened 2021-01-07
Move fallback redirection to its own controller

veganstraightedge:

n/a

- This PR moves the fallback route for redirects from the `pages#show` controller action to a new `redirects#show` controller and action.
- It deletes `pages#show` since it wasn't used anymore
- It moves `check_for_redirection` from a before action on **every** request to only getting called when no other route matches

I initially set out to move this for the organization of it. But in the end, I think it might could be a performance improvement because we're not doing 1-2 database reads for every request even ones that are a routes match. Instead, we're only making 1 database read when no other route matches.

- Go to a 404 route, like `/asdf`, expect the 404 page.
- Go to a normal route, like `/books`, expect it to work normally.
- Got to a redirect route, like `/j20`, expect it to redirect.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,6 @@ class ApplicationController < ActionController::Base
   before_action :set_current_theme
 
   before_action :set_site_locale
-  before_action :check_for_redirection
   before_action :strip_file_extension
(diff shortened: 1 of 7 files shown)


just1602 (approved):
LGTM

PR #1974, opened 2021-01-29
Lint all stylesheets

QueerSynd:

This is just linting all files in `the app/assets/stylesheets` folder. Nothing else has been changed. 

The website should look exactly as before 😄

### DIFF
--- a/app/assets/stylesheets/2017/base/_elements.scss
+++ b/app/assets/stylesheets/2017/base/_elements.scss
@@ -19,7 +19,15 @@ body {
   }
 }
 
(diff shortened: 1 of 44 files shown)


just1602:
Did you lint it with your editor or a standalone linter? 'Cause if you use a standalone linter, we could create a github action to lint the code and make sure the css is clean.

QueerSynd:
I limited myself to the IDE, but I thought about running a linting job, and I think it would be a good idea. I will look into it

just1602:
Yup, it would be perfect on another PR.

veganstraightedge (approved):

PR #1976, opened 2021-01-29
Move CSS/et al files from S3 to repo

veganstraightedge:


veganstraightedge:
There are some "invalid byte sequences" in many files that my Overcommit pre-commit hooks barfed on.

@QueerSynd @astronaut-wannabe Could you take a look at the files and see if there's hidden bad unicode characters or something?

veganstraightedge:
Nevermind, these are all already in `/public`.

https://github.com/crimethinc/website/tree/main/public/features

Maybe we need to move them from `public/` to `app/assets/stylesheets/`?
Then use the local ones instead of the CDN ones from the feature articles.

PR #1979, opened 2021-01-30
add linter config

QueerSynd:
Setting up a linter action for SCSS

### DIFF
--- a/.github/workflows/linter_scss.yml
+++ b/.github/workflows/linter_scss.yml
@@ -1,0 +1,17 @@
+  on: [push]
+
+name: Linter, SCSS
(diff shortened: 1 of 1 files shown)


just1602:
This look good to me, but I think we should merge the linting PR first.

QueerSynd:
> This look good to me, but I think we should merge the linting PR first.

Agreed, I got it to run with a standard-ish configuration, and there are a lot of errors of course 😄

just1602:
We could, but it would imply to add the style linter locally and if there is not a gem for the style linter it's a little bit more complicated.

PR #1990, opened 2021-02-13
Add `lang=foo` url param to the tools controller

astronaut-wannabe:

alt text

resolves `[link to github issue]`

This will allow you to filter the index of tools to a specific
language. `preferred_localization` still takes precedence with
deciding which translation to render, but the filter guarantees that
you will only be seeing tools that have as translation in your
filtered language

- go to review app
- see if filter works as expected


veganstraightedge:
This is pretty rad! Thanks for getting it going.

Do I have this right that the URL would look like this:

`http://localhost:3000/zines?filter[lang]=de` (locally, obvi)

If so, what's your thinking for doing `?filter[lang]=de` vs `?lang=de`

astronaut-wannabe:
> This is pretty rad! Thanks for getting it going.
> 
> Do I have this right that the URL would look like this:
> 
> `http://localhost:3000/zines?filter[lang]=de` (locally, obvi)
> 
> If so, what's your thinking for doing `?filter[lang]=de` vs `?lang=de`?

yup, you are correct. i'm not tied to any naming scheme for the param key. It was less thinking and more just my habit of defaulting to the json:api spec for stuff like this

app/controllers/tools_controller.rb
```diff
@@ -38,7 +38,24 @@ def set_body_id
   end
 
   def set_tools
-    @tools = tool_class.for_index(published_at: :desc).page(params[:page]).per(10)
+    @tools = tool_class.for_index(**filters).page(params[:page]).per(10)
+  end
+
+  def filters
+    {}.merge(sort)
+      .merge(language_filter).compact
```
veganstraightedge, line 46:
```suggestion
      .merge(language_filter)
      .compact
```

What do you think of this?

app/models/concerns/featureable.rb
```diff
@@ -7,8 +7,12 @@ module Featureable
...
 
   module ClassMethods
-    def for_index fallback_sort = { title: :asc }
-      where.not(hide_from_index: true).reorder(position: :asc).order(fallback_sort).english.published.live
+    def for_index fallback_sort: { title: :asc }, fallback_locale: 'en'
+      where.not(hide_from_index: true)
+           .reorder(position: :asc)
+           .order(fallback_sort)
+           .where(locale: fallback_locale)
+           .published.live
```
veganstraightedge, line 15:
```suggestion
           .published
           .live
```

Same thing here?

spec/system/tools_language_filter_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+describe 'filter[lang] for Tools Pages' do # rubocop:disable RSpec/DescribeClass
+  after(:all) { Zine.destroy_all } # rubocop:disable RSpec/BeforeAfterAll
+
+  before(:all) do # rubocop:disable RSpec/BeforeAfterAll
```
veganstraightedge, line 6: Will this get you the same thing without needing to disable the Rubocop rule?

```suggestion
  before(:context) do
```

spec/system/tools_language_filter_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+describe 'filter[lang] for Tools Pages' do # rubocop:disable RSpec/DescribeClass
```
veganstraightedge, line 3: Why does this system spec's `describe` need a special Rubocop comment, but these existing system specs don't?

https://github.com/crimethinc/website/tree/main/spec/system

astronaut-wannabe: @veganstraightedge https://github.com/crimethinc/website/blob/3407d59e9acddd2610e22a1d58fbb3b9990028af/.rubocop_todo.yml#L22-L30

veganstraightedge: Ooooh. I see.

spec/system/tools_language_filter_spec.rb
astronaut-wannabe: gave it a shot, makes no difference https://github.com/crimethinc/website/runs/1920963201?check_suite_focus=true

spec/system/tools_language_filter_spec.rb
astronaut-wannabe: i can put it there if it looks nicer. i honestly do not like that cop and would rather disable it

veganstraightedge: How about we split the difference?

What if we move that from TODO to `.rubocop.yml`, but rather than disabling it whole cloth, we just `Exclude:` that system spec folder?

astronaut-wannabe: sounds good to me

spec/system/tools_language_filter_spec.rb
veganstraightedge: Gotcha. Bummer.

veganstraightedge (approved):

PR #2205, opened 2021-11-01
[I18N] fix bug for displaying unicode urls

astronaut-wannabe:

alt text

We were getting the following errors for some international URLs:

```
  incompatible character encodings: ASCII-8BIT and UTF-8
```

You can reproduce this error using cURL:

```sh
  curl http://localhost:3000/languages/português-brasileiro
```

You cannot really reproduce this error by just pasting the above URL
into a browser's address bar because most modern browsers will
transparently URL encode Unicode

I found a good explanation and the fix on the [SO question][0]

[0]:https://stackoverflow.com/a/67400648

1. run ` curl https:/crimethinc.com/languages/português-brasileiro` and observe a `500`
2. run ` curl https://pipeline-to-fix-unicoe--bini3n.herokuapp.com/languages/português-brasileiro` and observe a `200`

### DIFF
--- a/app/views/2017/shared/_footer.html.erb
+++ b/app/views/2017/shared/_footer.html.erb
@@ -7,7 +7,13 @@
 
   <% if media_mode? && I18n.locale == :en %>
     <p class="site-mode-link-container">
-      <%= link_to(t("footer.site_mode"), url_for_current_path_with_subdomain(subdomain: :lite)) %>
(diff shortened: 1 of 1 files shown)


veganstraightedge (approved):
Looks good. Would you mind extracting that out to a helper method please? Either here or in  another PR. 👍🏻

PR #2258, opened 2021-11-28
Revert "Return `404` when archive path is incorrect"

### LINKED ISSUE #2257: Pagination links aren't working
From any page with a collection of articles that spans more than one page.
We expect that the links in the pagination would have `/page/2`, `/page/3`, etc.
But none of them do.

# Development
http://localhost:3000/2000/page/2

# Production 
https://crimethinc.com/2021/page/2

astronaut-wannabe:
Reverts crimethinc/website#2200

resolves #2257


veganstraightedge (approved):
Thanks!

PR #2377, opened 2022-01-26
Bump ruby version to ruby 3.1

just1602:

excited person

It upgrades ruby to 3.1.0

* Check that all the tests pass
* Check on the staging environment that everything seems to work


just1602:
@veganstraightedge @astronaut-wannabe 

I'm not sure I like the new default that avoid the value in a hash if it has the same name has the key. Do you prefer I fix them or I had an exception to the rubocop config?

veganstraightedge:
> @veganstraightedge @astronaut-wannabe
> 
> I'm not sure I like the new default that avoid the value in a hash if it has the same name has the key. Do you prefer I fix them or I had an exception to the rubocop config?

Good question! Can you point to an example of this so I can better understand, please?

veganstraightedge:
Oh I see now. In the linter/rubocop PR check failure output.

Yeah, that's fucking weird. I propose that we disable that rule and move it into `.rubocop.yml`. Maybe in the future our brains will have adjusted to like it better, but for now, it's weird to my brain.

just1602:
@veganstraightedge  do you think it could be possible to have a staging deployment to validate it works as expected?

veganstraightedge (approved):

PR #2417, opened 2022-02-16
Add #notes to Article, re-organized article admin form

veganstraightedge:
This PR adds a `notes` field to the Article model and admin form.

It also re-arranges the admin/articles form to keep `notes` next to `content`.

The notes and content fields get combined by the `Article#content_and_notes` method which is what is used by `Article#content_rendered` method.

image

### DIFF
--- a/app/controllers/admin/articles_controller.rb
+++ b/app/controllers/admin/articles_controller.rb
@@ -141,7 +141,7 @@ module Admin
 
     def article_params
       permitted_params = params.require(:article).permit(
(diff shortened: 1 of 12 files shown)


just1602:
Is there any reason we want to split the notes from the aritcles?

just1602 (approved):

veganstraightedge:
@just1602 It's setup for a design change/feature that I have in the works for a redesign of the site.

Issue #47, opened 2016-08-31
Syndication: Twitter


veganstraightedge:
- Auto tweet link to article at publish time
- Smart truncation / elipsing
- Field in Article form for tweet preview with character count
- Ensure Twitter cards work with proper meta tags
- Consider @adactio's post: https://adactio.com/journal/9881
- [twitter tweet api doc ](https://developer.twitter.com/en/docs/tweets/post-and-engage/api-reference/post-statuses-update)
- [twitter upload media api doc](https://developer.twitter.com/en/docs/media/upload-media/overview)
- Maybe use [this library](https://github.com/sferik/twitter)

just1602:
We've been banned.

PR #2426, opened 2022-02-21
Add Tweet class and rake task

veganstraightedge:
Building upon https://github.com/crimethinc/website/pull/2357…

This PR adds a `Tweet` class that uses the `twitter` gem to publish to the Twitter API.

# Usage

```ruby
# It works for any of these tools: Book Logo Poster Sticker Video Zine
# For example…
tool = Book.first 

Tweet.for(tool).publish
```

In the `rake` task, `Tweet.for` is given a random tool from the `RandomTool` class.

```ruby
Tweet.for(RandomTool.sample).publish
```

# Next

And then it tweets it, assuming the four relevant environment variables have valid values.

After the environment variables are in place, we can turn on an hourly job using the [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler).


app/models/tweet.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def image_file
+    Down.download image_url
+  end
+
+  def thing_url
+    "https://cwc.im#{thing.path}"
+  end
+
+  def twitter_client
```
just1602, line 43: Later we may want to extract the twitter client to a class if we want to automatize the syndication of article on twitter, but it's really clean!

veganstraightedge: That's a great idea, @just1602. Let's definitely do that.

It's one of the oldest issues even!
https://github.com/crimethinc/website/issues/47

PR #2443, opened 2022-02-28
Allow unmetered access sometimes

veganstraightedge:
***


just1602 (dismissed):

just1602:
Sounds good, it's a nice way to allow some tool we know to not be throttle!

PR #2444, opened 2022-02-28
Update sitemap.rb

just1602:

I don't know why, but sitemap.xml is not redirected by our middleware. I
thought I'd ahve to add `|| @req.path == '/sitemap.xml'` to the
`exit_early?` method, but it's not the case.

To validate I add the `public/sitemap.xml` file and I validate the I got
the right content when I query `/sitemap.xml` and it works. The only
things to keep in mind is since I create a dummy `public/sitemap.xml`
file, we need to regenerate one every time we deploy the application.

To ensure the sitemap is generate, can I run the sitemap:clean and 
sitemap:refresh rake task in the postdeploy section of the app.json file?

pinging /sitemap.xml shouldn't give you a 404.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -70,7 +70,7 @@ SitemapGenerator::Sitemap.create(default_host: 'https://crimethinc.com', compres
     add "/#{year}/"
   end
 
-  [Book, Episode, Page, Podcast, Video, Tools].each do |model|
+  [Book, Episode, Page, Podcast, Video, Zine, Journal, Issue, Episode, Poster, Sticker, Logo].each do |model|
     model.find_each do |page|
       add page.path, lastmod: page.updated_at
     end
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -27,22 +27,28 @@ SitemapGenerator::Sitemap.create(default_host: 'https://crimethinc.com', compres
   static_paths = [
     '/about/',
     '/arts/submission-guidelines',
-    '/books/into-libraries/', '/books/lit-kit/',
+    '/books/into-libraries/',
+    '/books/lit-kit/',
     '/categories/',
     '/faq/',
+    'games/j20',
     '/get/',
     '/kickstarter/2017/',
     '/listen/',
     '/library/',
     '/read/',
-    '/rt/', '/rt/archives/',
+    '/rt/',
+    '/rt/archives/',
     '/start/',
-    '/store/', '/store/audio/', '/store/added/',
+    '/steal-something-from-work-day',
+    '/store/',
+    '/store/audio/',
+    '/store/added/',
     '/tce/',
     '/tools/',
     '/watch/'
   ]
-  tce_languages = %w[czech deutsch espanol polski portugues quebecois slovenscina slovensko]
+  tce_languages = %w[czech deutsch espanol espanol-america-latina polski portugues quebecois slovenscina slovensko فارسی 日本語 한국어 lietuvos portugues ภาษาไทย]
 
   tce_languages.each do |lang|
     static_paths.push("/tce/#{lang}/", "/tce/#{lang}/get/")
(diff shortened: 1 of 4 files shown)


config/sitemap.rb
```diff
@@ -60,11 +66,11 @@
     add "/categories/#{category.slug}/"
   end
 
-  [2017, 2016, 2015, 2014, 2013, 2012, 2011, 2010, 2009, 2008, 2007, 2006, 2005, 2004, 2003, 2002, 2001, 2000, 1997, 1996].each do |year|
+  [1996..Time.zone.today.year].to_a.each do |year|
     add "/#{year}/"
   end
 
-  [Book, Episode, Page, Podcast, Video].each do |model|
+  [Book, Episode, Page, Podcast, Video, Tools].each do |model|
```
veganstraightedge, line 73: Have you confirmed that this works? 

I wouldn't expect it to:

```ruby
irb(main):001:0> Tools
(irb):1:in `<main>': uninitialized constant Tools (NameError)
```

Instead, I think what we need here is to list out each tool model (in its singular form):
```ruby
Zine
Journal
Issue
Episode
Poster
Sticker
Logo
```

Note: we have "Music" on `/tools`, but it's just a redirect to the band camp page. So we don't need to include it in sitemap.

public/sitemap.xml
```diff
@@ -0,0 +1 @@
+<sitemap>test</sitemap>
```
veganstraightedge, line 1: Let's see if we can find a way to not need this.

spec/requests/clean_paths_spec.rb
```diff
@@ -12,4 +12,12 @@
...
     expect(response.header['Location']).to eq('/about/faq?test=true')
   end
+
+  it 'do not strip .xml extension for sitemap.xml' do
+    get 'http://example.com/sitemap.xml'
+
+    expect(response.status).to eq(200)
+    expect(response.header['Location']).to eq(nil)
+    expect(response.body).to eq('<sitemap>test</sitemap>')
+  end
```
veganstraightedge, line 22: Maybe instead, we need to run the rake tasks that you mention in the PR description.

veganstraightedge:
>  we need to regenerate one every time we deploy the application.
> To ensure the sitemap is generate, can I run the sitemap:clean and
> sitemap:refresh rake task in the postdeploy section of the app.json file?

Good thinking!

1. `app.json` doesn't get used on each production deploy. It only gets used when creating a new app. So, in our case, that means only the review apps that our Heroku pipeline makes for each PR.
2. However, the `release` stage in `Procfile` gets used on every deploy. That's how our migrations get run.
3. Finally, let's keep this separated as two PRs: 
  - This one, focused on getting a good working `sitemap.xml` (that you manually run the rake tasks to create)
  - A subsequent PR to figure out the auto-generated process (on each deploy?)

Great work @just1602! 🎉

config/sitemap.rb
just1602: Done.

spec/requests/clean_paths_spec.rb
just1602: This is pretty clean and it make me realize that the `clean_paths` middleware does not remove the extension since the sitemap is compressed and the extension is `.gz` and not `.xml`.

spec/requests/clean_paths_spec.rb
```diff
@@ -12,4 +14,16 @@
 
     expect(response.header['Location']).to eq('/about/faq?test=true')
   end
+
+  it 'do not strip .xml extension for sitemap.xml' do
```
veganstraightedge, line 18:
```suggestion
  it 'does not strip .xml extension from sitemap.xml' do
```

veganstraightedge (approved):
I ran the `sitemap:refresh` task and it failed because `Issue` didn't have `updated_at` (or `created_at`). So I added them in a #2459 then rebase this PR to have them.

Great job! 🎉

PR #2461, opened 2022-03-05
[I18n] Add GL / Galician / Gallego locale configuration

veganstraightedge:
https://en.wikipedia.org/wiki/Galician_language


just1602:
Did someone translate an article in Gallego?

PR #2476, opened 2022-03-11
[#2475] Delete files from the old online store

### LINKED ISSUE #2475: Delete public/store
Since the store has been refactored do we need to keep the `public/store` folder or we can remove it?

just1602:

alt text

resolves #2475

It remove the files of the old store from the `public/` folder.

Check if once those files removed, nothings break in the actual store, but it shouldn't be the case since all the assets are on shopify cdn.


veganstraightedge (changes requested):
`public/store-maintenance/index.html` 

We need to keep this file. 

We use whenever we need to take the store offline for maintenance.

Otherwise this PR looks good.

just1602:
Should be alright now!

veganstraightedge (approved):

Issue #2345, opened 2022-01-11
Allow people to download pdf version of an article


just1602:
I know for some articles there's a zine version and I think it's really nice that way, but since there's not a zine version of every article it would be nice to have the option to download an article as pdf. It's not a way to replace zine version at all, but it's more because sometime you want to be able to read an article while not having internet, save it in your zotero, mandelay, onenote, etc library or simply annotate it. Personally, I use wallabag to save and tag articles, but I know people or store pdfs locally and could be interest by that kind of feature.

We could have a template :

* with a "cover page" with basic information about crimethinc. and the article
* without menus, newsletter subscription form ,etc

The other option would be to have a print specific css style that would hide a bunch of things and rework how we display te content.

PR #2480, opened 2022-03-14
Add route and admin button to download .docx of article

veganstraightedge:

Adds a route and a button (in admin/authed site header) to download a `.docx` version of an Article.
It has the `[[ … ]]` embeds stripped out of the content too.

This was requested by zine makers because importing a `.docx` into InDesign is much easier that copy/pasting the Markdown or HTML.

Screen Shot 2022-03-14 at 12 24 39 AM


just1602:
This is really nice and I think I could do the same thing for this issue #2345 !

PR #2497, opened 2022-03-21
Update badge svg url to hvae the real status

just1602:
The GitHub badge were always saying our tests were failing so I update the url to point on the right svg. 

According to [the documentation](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) it seems we can't point to a series of workflow, but only on specific ones, so I put the badges for system and non-system tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@ Ruby on Rails app that powers https://crimethinc.com
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


astronaut-wannabe (approved):

astronaut-wannabe:
this is possibly fixed by https://github.com/crimethinc/website/pull/2496 and we could not change the label, that said I'd like to see at a glance which parts of CI are failing so I still think this change is a good one even if you fixed the original label via https://github.com/crimethinc/website/pull/2496

PR #2501, opened 2022-03-22
Add a workflow to automerge dependabot PRs

### LINKED ISSUE #2500: Automerge dependabot pull requests
Since we have a lot of pull request coming from dependabot and a test suite we can kind of trust, we should merge dependabot PR when buld are green.

We should inspire ourselves from [this example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/.github/workflows/build.yml#L154-L168)

**EDIT**

After some duckduckgoin maybe we should simply use [this action](https://github.com/marketplace/actions/dependabot-auto-merge)

just1602:
If we want to do it all by ourselves, we should have an action that would run all the other tasks on the PR and when everything is passed, we could run the command to merge the PR automatically. Instead, I found an actions that do it for us, so we'll try it and if it doesn't work, then I'll write something to orchestrate and automerge dependabot stuff.

Fixes #2500 

**EDIT**

I add a if check to only run the action on PRs made by the dependabot bot, otherwise it emits a warning.

### DIFF
--- a/.github/workflows/auto-merge-dependabot.yml
+++ b/.github/workflows/auto-merge-dependabot.yml
@@ -1,0 +1,15 @@
+name: Automerge dependabot pull request
+
+on:
+  pull_request:
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2.3.1
+        with:
+          target: minor
+          github-token: ${{ secrets.GITHUB_TOKEN }}


astronaut-wannabe (approved):
code-wise this LGTM.

With regards to the feature itself, I don't really think it is a good idea (not just because of the security concerns mentioned on the issue)

However, my comment on the issue grew to over 500 words, which seemed a bit excessive :laughing: If this improves other's workflows I can adjust to it

the short version of my argument is that:

1. I don't consider dependabot PRs as chores, rather as a newsletter tailored to our specific dependencies
2. I'd rather have my mental model of a system up to date and dependabot PRs pile up, as opposed to my mental model having increasing uncertainty but more up-to-date gems
3. the links to the commits in a dependabot PR are a valuable resource for learning engineering principles from the open source community

just1602:
To be honest, the security arguments kind of speak to me and I tend to read commit logs and changelog to avoid bad surprises and see what is going to be merged in the project.

The answer to so much gem update is probably that we use too much gems and should reduce our number of dependencies and / or gems are updated too quickly and stuff like rubocop and aws-core-sdk should probably slow down their release pace.

As I said in the issue maybe the solution would be to auto-approbe those issue to avoid all the back and forth between the different PR tabs to approve then merge the PRs.

PR #2503, opened 2022-03-23
[BUG] Update locale.articles_count when an article is destroyed

astronaut-wannabe:

alt text

The two bugs are:

- If an article is destroyed, we never trigger the callback. This
  doesn't have a huge effect on locale's with a lot of articles, as
  eventually the value is updated. However, locales with infrequent
  articles can get into a bad state and stay there for awhile

- If an article has its locale changed, the original locale is never
  updated, only the new locale it has been changed to.

[0]:https://crimethinc.com/languages

rspec tested in this case

We have an `after_save` hook that will look up an article's locale,
and update the `Locale.articles_count` value for the article's
associated locale.

This value is used in rendering the [languages page][0] on the site.

We encountered a situation where we had 0 Galician articles in the
database, yet the website was displaying a `1` for Galician on the
site.

After looking into the code we realized that our `after_save` hook
only covered the use case of an article being created or updated, not
destroyed, and that there was also a bug with the `update` case.

- [X] Does the code you updated have tests? If not, could you add some please?
- [X] nope ~Does this pull request require any new server dependencies which need to be added to the build process? If so, please explain what.~

# Acceptance Criteria

- [X] This pull request does not cause the database export script to become out of sync with the db schema

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -21,7 +21,8 @@ class Article < ApplicationRecord
   validates :summary, length: { maximum: 200 }
 
   before_save :update_or_create_redirect
-  after_save  :update_locale_articles_count
+  after_destroy :update_locale_articles_count
+  after_save :update_locale_articles_count
 
   default_scope { order(published_at: :desc) }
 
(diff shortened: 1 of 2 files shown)


app/models/article.rb
```diff
@@ -126,13 +127,16 @@ def update_or_create_redirect
...
+    article_locales = Locale.where(abbreviation: [locale, original_locale].compact)
 
-    return if article_locale.blank?
+    return if article_locales.blank?
 
-    articles_count = Article.live.published.where(locale: article_locale.abbreviation).count
-    article_locale.update(articles_count: articles_count)
+    article_locales.each do |article_locale|
+      articles_count = Article.live.published.where(locale: article_locale.abbreviation).count
+      article_locale.update(articles_count: articles_count)
```
just1602, line 139: Is there a reason why we keep computed data in table instead of computing them on the fly? It'd avoid all this update / delete logic, tho?

just1602 (approved):

app/models/article.rb
astronaut-wannabe: I assume it is because there would be a large number of database calls on the /languages page. perhaps partials could be cached at the view layer, but that is a larger refactor and I am not sure it would even be a better situation.

I lean towards moving this logic into a background job, that has a unique constraint, to isolate the logic and testing

just1602: Ok, for now let's merge it that way, but let's move it to a background job or find a way to avoid this logic in another PR.

Issue #1448, opened 2020-01-06
Add translation / localization to admin forms (for Spanish translations)


veganstraightedge:
We were given a handful of translations of form labels for the article admin form. This would be helpful for content folx working on Spanish articles. However, the admin forms aren't set up to use the locale YAMLs. 

# Step 1

- [x] Get the current English forms using the `en.yml` / `en/admin.yml` files.

# Step 2

- [x] Replace the English values with Spanish ones for the following:

English | Spanish
------------ | -------------
| Title | Título |
| Subtitle | Subtítulo |
| Content | Contenido |
| Tags | Tags |
| Categories | Categorías |
| Comma separated. For example, dogs, cats, etc. | Separados con comas. Por ejemplo: perros, gatos, etc. |

PR #2510, opened 2022-03-26
Issue #1448: spanish translation of admin form labels

fluttermouse:
I've translated the english labels for the admin form to spanish (in es.yml). In addition to the form labels mentioned in the issue, I also translated the "search_results" label.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -40,7 +40,7 @@ es:
 
   search:
     form_placeholder: Busca en los archivos…
-    search_results: 'Resultados de la búsqueda de: %{query}'
+    search_results: 'Resultados de la búsqueda: %{query}'
     advanced:
       heading: Búsqueda avanzada
       form:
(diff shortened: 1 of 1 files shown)


config/locales/es.yml
```diff
@@ -40,17 +40,17 @@ es:
 
   search:
     form_placeholder: Busca en los archivos…
-    search_results: 'Search results for: %{query}'
+    search_results: 'Resultados de la búsqueda de: %{query}'
```
astronaut-wannabe, line 43: would the second "de" change based on the value of `%{query}`?

If so, maybe we just translate "Search results:" rather than "Search results for:"? What do you think?

astronaut-wannabe (approved):
this lgtm. one translation question on the search results label

config/locales/es.yml
fluttermouse: I've updated "search_results" to "Resultados de la búsqueda: %{query}".

astronaut-wannabe (approved):

astronaut-wannabe:
resolves #2510

astronaut-wannabe:
Thanks for the PR! @fluttermouse

PR #2513, opened 2022-03-27
Migrate to Stimulus and import maps for JS

astronaut-wannabe:
## commit d914b6efc446001a099f1554525099e937be11bd

Install importmap-rails

As a first step towards migrating to the new Rails 7 default of
hotwire JS via import maps, i have followed the installation
instructions listed [here][0]

1. Add `importmap-rails` to your Gemfile with gem 'importmap-rails'
2. Run `./bin/bundle install`
3. Run `./bin/rails importmap:install`

[0]:https://github.com/rails/importmap-rails/blob/86970ab8f21c8842e86bee4fc887d3bc53fa9848/README.md#installation

## commit 04c748b70355b3dbacee27c7b9f502e89bafa551

Begin wiring in the importmaps

We will preload the `app/javascript/application.js` and any modules
imported there (eventually the stimulus framework files will be
included)

Any custom imports for specific pages can be included by wrapping in a
`content_for` block (which is yielded to in the layout's head tag) as
described in the docs [here][0]

[0]:https://github.com/rails/importmap-rails/blob/86970ab8f21c8842e86bee4fc887d3bc53fa9848/README.md#selectively-importing-modules

## commit 247dd7e8b3535e5638916ecf36c66d962c173524

Install stimulus JS with import maps

I installed stimulus with the installer as described in the [docs][0]

The intent is to move our few JS features over to stimulus
controllers:

- the live blog poller
- the admin character counter

This brings us more in line with the default rails app setup and will
make following hotwire and other rails guides easier in the future

### removing `jquery-rails`

I also removed the jquery rails gem because I don't think we need it
for anything:

- none of our custom JS uses jquery
- jquery is optional for bootstrap as I think they migrated to vanilla
  JS APIs in the latest versios

[0]:https://github.com/hotwired/stimulus-rails/blob/0849a02be09e91f99d8a4dd85ea328ce6691f3c5/README.md#installing-with-installer

## commit 15aa6e238e6fb34cb80d63a5b9b429602cfa98dc

Migrate admin text box counter to stimulus

I migrated the text counter (used for summary text and tweet length
text) to a basic Stimulus controller

The basic functionality is this:

- When you type, it triggers `count()`, which updateds the current
  character count (`currentValue`)
- If the character count exceeds `maxLengthValue`, a bootstrap
  `invalid` css class is applied to the input
- The `invalid` class is removed from the input once the count goes
  down

We do not actually block typing passed maxLength in the input to make
the editing of a tweet easier

## commit f4fd6773e26faad0d41b3847106b3eb5c9cf9249

Migrate live blog polling to Stimulus

I migrated the live blog polling logic to stimulus. In the original
code there was a queueing mechanism set up, I think with the hope that
eventually we might consume messages off of the queue and actually
update the page's content without a refresh

That never happened and all it actually does is pop up a banner that
alerts the user to refresh their page to see new posts. That is the
featureset I migrated to stimulus.

The live updating can more easily be accomplished with hotwire now
that that is an option, so perhaps we will revist this feature going
forward. In the meantime the live blog functionality as it stands with
this PR is:

- If an article is a collection, this live_blog controller is loaded
  with the UI in a hidden state
- On page load, the current UTC time is stored in the
  controler (`pageLoadedAt`)
- in the `initialize()` hook, the poller is started with a 1 minute
  interval
- I modified the collection endpoint to accept the parent articles
  slug and the `pageLoadedAt` utc value
- The endpoint queries for any `article.collection_posts` published
  after `pageLoadedAt`
- when the controller receives the response, it just updates the
  `updateCount` value with the length of the returned posts array
- if the value changes and is greater than 0, the onchange callback
  updates the banners text and makes it visible

A note on the changes in the collection endpoint: I left the original
codepath in place just in case someone else was consuming this
endpoint, but I think that original path is now dead code. Maybe when
we revisit this feature we can clean that up

## commit b45cd358e823f2a74f425131efeff332d3179626

migrate support.js to importmaps

I do think the support page slider is a good candidate for a stimulus
controller.

That said, I do not know how to set stripe up for testing so I decided
to not modify any of that javascript. Instead I am using this as an
example of how to selectivley load some custom JS on specific pages
with import maps

1. I moved the file from the `app/assets/javascript` directory to the
   `app/javascript` directory as this is where JS will be placed for
   import maps
2. I removed the file from the asset pipeline manifest (I think this
   is correct, but I need to double check that)
3. I used the `content_for :head` to load the JS selectivley in
   `/support` views

I checked that the JS file does indeed load on the `/support` change
with these changes (at least in development)

## commit 346ce3a0f84b987bf12443b8a02fb6e7cf062a10

load bootstrap and popper JS in admin and SSFWD pages

I couldn't actually find an example of where bootstrap JS is used in
admin, but the Gemfile asserted it was being used for something.

So, I moved bootstrap and the steal-something-from-work javascripts in
a similair way that I did the support.js

## commit 205b1323780e5998d88331122ce730a68ec1b131

PR Feedback

Co-authored-by: Shane Becker <veganstraightedge@gmail.com>


astronaut-wannabe:
demo app: https://pipeline-to-migrate-to--sr9qtj.herokuapp.com

just1602 (approved):

app/views/2017/articles/_live_blog.html.erb
```diff
@@ -0,0 +1,11 @@
+<div id="floating-alert"
+     class="alert alert-floating alert-notice"
+     hidden=true
```
veganstraightedge, line 3:
```suggestion
     hidden="true"
```

app/views/2017/articles/_live_blog.html.erb
```diff
@@ -0,0 +1,11 @@
...
+     data-controller="live-blog"
+     data-live-blog-page-loaded-at="<%= Time.current.utc.to_i %>"
+     data-live-blog-url-value="<%= "/articles/#{article_slug}/collection_posts" %>"
+     data-live-blog-target="container">
+  <a class="close"
+     data-action="live-blog#close"> &#10006;</a>
```
veganstraightedge, line 9:
```suggestion
     data-action="live-blog#close"> ✖</a>
```

app/views/admin/articles/form/_syndication.html.erb
```diff
@@ -2,20 +2,32 @@
...
+         data-admin--text-box-counter-max-length-value="250"
+         data-admin--text-box-counter-error-class="is-invalid"
+         data-controller='admin--text-box-counter'>
       <%= form.label :tweet, class: "form-label" %>
-      <%= form.text_area :tweet, class: "form-control form-control-lg", data: { 'max-length': '250', 'feedback-box': 'tweet-length' }, rows: 3 %>
-      <div id="tweet-length">0</div>
+      <%= form.text_area :tweet, class: "form-control form-control-lg",
+          'data-admin--text-box-counter-target': 'source',
+          'data-action': 'admin--text-box-counter#count',
+          rows: 3 %>
```
veganstraightedge, line 13:
```suggestion
      <%= form.text_area :tweet, 
          class: "form-control form-control-lg",
          "data-admin--text-box-counter-target": "source",
          "data-action": "admin--text-box-counter#count",
          rows: 3 %>
```

app/views/layouts/2017/_application.html.erb
```diff
@@ -6,6 +6,8 @@
              prefix: html_prefix do %>
   <head>
     <%= render_themed "shared/head" %>
+    <%= javascript_importmap_tags %>
+    <%= yield(:head) %>
```
veganstraightedge, line 10:
```suggestion
    <%= yield :head %>
```

app/views/layouts/2020/_application.html.erb
```diff
@@ -8,7 +8,9 @@
     <%= stylesheet_link_tag Current.theme, media: "all" %>
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Markazi+Text:400,500,600,700&display=swap", media: "all" %>
     <%= render_themed "shared/translation_missing_css" %>
-  </head>
+    <%= javascript_importmap_tags %>
+    <%= yield(:head) %>
```
veganstraightedge, line 12:
```suggestion
    <%= yield :head %>
```

veganstraightedge (dismissed):
Looks great! Thanks for taking this on. 🎉

All of my comments/code suggestions and non-blocker syntaxy nit picks. You may feel free to ignore them all in this PR if you want.

astronaut-wannabe:
needs a new :heavy_check_mark:  after merging in the changes

just1602 (approved):

PR #2616, opened 2022-05-30
Use setup-ruby action builtin cache

just1602:

alt text

It removes some step of the github action workflow to use the setup-ruby action builtin cache

Check that the pipelines are still working and that the cache is working

I was working on another rails project CI and find this possibility while comparing with the Crimethinc. CI setup.

# Acceptance Criteria


just1602:
In my next PR, I'll enable erblinter since there's some fix to do to make it pass.

PR #2918, opened 2022-11-29
Use :rack_test as default capybara driver

just1602:
The capybara documentation was suggesting that if the app was working without JavaScript we could use the `:rack_test` driver which is smaller and faster, and simply configure a `Capybara.javascript_driver` and use the `js: true` annotation on specs that need to use the `javasdript_driver`.

It's not a lot faster, but on my computer it's 4-5 sec faster, which is not bad, tho.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -18,4 +18,4 @@ RSpec.configure do |config|
   config.filter_rails_from_backtrace!
 end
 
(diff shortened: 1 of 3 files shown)


veganstraightedge:
I like this! Great job.

veganstraightedge (approved):

PR #2962, opened 2022-12-14
Fix publication date alignment on home feed entry

just1602:

alt text

Nope!

The publication date was sliding just under the `h-entry` card. This PR realigns it. Sadly, it uses `position: absolute;` instead of using flex, but I guess that my time would be best invested in working on the new design then refactoring this one.

The actual home feed with publication date that slide under the card.

image

The home feed after aplying this PR :

image

It also works on mobile:

Screen Shot 2022-12-13 at 20 21 12

### DIFF
--- a/app/assets/stylesheets/2017/views/_home.scss
+++ b/app/assets/stylesheets/2017/views/_home.scss
@@ -90,6 +90,12 @@
       }
     }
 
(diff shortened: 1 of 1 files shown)


veganstraightedge (approved):
thanks!

PR #2963, opened 2022-12-14
Remove macos-strap from README since the URL give a 404

just1602:
The actual URL of the `strap.sh` script return a `404`, so I'd remove it. When I checked on Google, I found [this site](https://daptiv-macos-strap.herokuapp.com/), I don't know if it's the right thing. If so, then I will just change the URL, if it's not I'll just remove ti.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ Ruby on Rails app that powers https://crimethinc.com
 
 ## Short Version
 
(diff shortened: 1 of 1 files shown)


veganstraightedge:
https://github.com/MikeMcQuaid/strap#usage

Looks like it moved (after Heroku ended free plans, probably)

https://strap.mikemcquaid.com

veganstraightedge (approved):

PR #2992, opened 2022-12-30
Add IG to social icons

veganstraightedge:
📸


just1602:
I can't beleive instagram is so popular that people are starting over for a second time, tho.

Issue #1823, opened 2020-10-24
Dockerize deployable site


veganstraightedge:
We deploy to and run on Heroku.
That's still the primary move.

But if we have to leave Heroku in some future, we'd like to be able to do that easily and reliably. Docker seems like answer for that.

Heroku even has a Docker story these days. Last time I looked (a year ago), not all of the normal Heroku features worked in Docker offering. It's worth exploring though.

# Heroku + Docker 
https://devcenter.heroku.com/categories/deploying-with-docker
https://www.heroku.com/deploy-with-docker
https://devcenter.heroku.com/articles/container-registry-and-runtime

# Non-Heroku

Ultimately, this is about being able to deploy and run the site from a non-Heroku host. That means: AWS, GCP, Azure. But also, homegrown servers. Ideally, both situations would be accommodated by this.

junkyard-prog:
Hey is it cool to just grab an issue and go or do I need to ask to be assigned and such?

just1602:
Just assign yourself and go for it :D

PR #3028, opened 2023-01-17
started working on dockerization

jessehorne:

do it live

resolves [Issue 1823](https://github.com/crimethinc/website/issues/1823)

This PR introduces three new files and some minor modifications which I believed necessary to a script or two...
1) `Dockerfile` which contains the image definition for the rails service itself
2) `docker-compose.yml` which contains the compose yaml for launching the rails service *and* the postgres instance separately.
3) `config/environments/docker.rb` which is basically a clone of `development.rb` but `database.yml` specifies the username for the database to be `postgres`. This was the most straight forward way to do this without having to do complicated changes on dockers side.

Another thing to note, if you see `docker-compose.yml` you'll see an environment variable specified named `POSTGRES_HOST_AUTH_METHOD` which is set to `trust`. This just makes it so we don't need to auth if we're connecting locally, from what I understand, but I truly don't fully understand this flag in its entirety so we probably shouldn't deploy this until we make sure we know what we're doing here. :P

I've added instructions on how to use this in the README but basically you just need to run the following two commands and it *should* work, considering any previous setup necessary depending on your distro/os.

```sh
docker build -t crimethinc .
docker compose up
```

p.s

The database and rails service being separate is necessary. Docker containers are designed to run one thing at a time. It's not impossible to run the database and rails service together but it is not a good idea. :D

Depends on how we deploy but basically, everything should be tested as normal, manually. Probably scrape the site and make sure everything is accessible just the same. Should be straight forward. We can probably test it however you have tested it before.

See above. There's a bit to consider here but I got it working to its current state and looking forward to discussing what we could do to make it better.

# Acceptance Criteria


Dockerfile
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1
+FROM ruby:3.1.3
+WORKDIR /crimethinc
+COPY . .
+RUN apt-get update -qq && apt-get install -y nodejs postgresql-client ruby-full git curl libssl-dev libreadline-dev zlib1g-dev autoconf bison build-essential libyaml-dev libreadline-dev libncurses5-dev libffi-dev libgdbm-dev
```
just1602, line 5: You probably don't need to install `ruby-full` since the base image is `ruby:3.1.3`. In fact, I'd double-check if you don't only need `nodejs` and `postgresql-client`.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN gem install bundler --conservative && gem install os && bundle install
+
+CMD bundle exec rails db:create && bundle exec rails db:migrate && ./script/server
```
just1602, line 9: You probably don't want to run `db:create` each time you start the application. You should probably provide an entrypoint that point to `bin/rails` and set the command to `bin/rails server`. Since the entrypoint is gonna be `bin/rails`, we're gonna be able to do stuff like `docker run crimethinc:latest db:create db:migrate`.

Also, you can run multiple task as once, so you could do `bundle exec db:create db:migrate` in one command.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
+services:
+  db:
+    image: postgres
```
just1602, line 4: You probably want to specify a version of the postgres image since data format are not necessary compatible between version, so if you redeploy and a new version has been release it may not boot because the format have change, and we need to migrate the data first.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    volumes:
+      - ./tmp/db:/var/lib/postgresql/data
```
just1602, line 6: I don't know if there's a way to specify this volume at runtime or set it to a more generic path, but we don't want production data to be keep in a `tmp` directory. The best place would probably be something like `~/crimethinc/data` where assuming crimethinc has its own user on the host machine.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
just1602, line 8: It's how I setup my postgres database in local, but I don't know the best practice related to production and security. Myabe it's better to have a separate password for the database.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    ports:
+      - 5432:5432
+  web:
+    image: crimethinc:latest
+    ports:
+      - "3000:3000"
```
just1602, line 14: Here, if you don't setup another docker-compose `service` to front the Crimethinc. Web app like `nginx`, `haproxy` or anything else that can be used as a reverse proxy, you'll need to expose the port `3000` to the port `443` of the host machine. But if you do so, it'll also mean you need to handle SSL certificate.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    links:
+      - db
+    environment:
+      - RAILS_ENV=docker
```
just1602, line 18: For a first draft, this is gonna be fine, tho. But I think we really need to find a way to run every environment in docker. You should probably check this [PR](https://github.com/rails/rails/pull/46762). In rails 7.1 that is gonna be release soon, DHH add the possibility to generate a default `Dockerfile`.

config/environments/docker.rb
```diff
@@ -0,0 +1,73 @@
+require 'active_support/core_ext/integer/time'
```
just1602, line 1: As mention later in another comment, we should probably find a way to run every environment in the docker container.

just1602 (commented):
Hey! Thanks for this PR, this is really nice! :tada: 

I've added some comments, don't hesitate if you have any questions.

Dockerfile
jessehorne: Oh hey, good catch! I'm really not sure how I ended up with all of those dependencies. Must have been while I was tinkering around trying to get this to work. You were right, we only need `nodejs` and `postgresql-client` apparently.

Dockerfile
jessehorne: This was great advice. It ended up being so cleaner and more usable.

veganstraightedge:
@jessehorne I'm so sorry that this PR has sat here for so long with no real response from me.  (the past several months have been distracted) Thank you so much for putting in this effort on the project. 🎉

I'm gonna close this PR at this time though. 

This issue is only looking for a Docker setup that can be deployed. Not a setup for Docker based development. 

I'm still open to a PR that is scoped to just a deployable Dockerfile though. 🙂

PR #3312, opened 2023-06-16
Remove twitter deps

just1602:
This PR also add `ruby-lsp` as a dev deps since it's really easier for me to have it inside the project and not system-wide.

The main goal of this PR is to remove the dependency to the `twitter` gem. I also remove the `Twitter` model since it was the class using the `TwitterClient`. I was wondering if you want to also remove the `RandomTools` class? I didn't remove it for now since it wasn't coupled to the `Twitter` model, but we don't use it anywhere else, so unless you want to create the same kind of bot, but for mastodon, I'd also remove the `RandomTools` class.


veganstraightedge (approved):
Great thinking, @just1602! 

I like the idea of us creating a Random Tools Mastodon toot-bot. So, maybe leave the RandomTool class around for now. 👍🏻

veganstraightedge:
Note to our future selves: 

Revive the `Tweet` class as a `Toot` class when we build Mastodon syndication.

PR #3511, opened 2023-09-06
Add auto-merge for dependabot PR except when it's a major upgrade

just1602:
I give up, there's to many dependencies updates. There'll always be a risk of supply chain attack, even if we review dependencies updates manually. The key will probably be to keep the number of dependencies as low as possible, but I wanna try auto-merge to see if it reduces the amount of work.


veganstraightedge:
i like it, but how about we start the experiment with targeting just patch/tiny version changes?

then when we have confidence in the system, we move it up to minor

Issue #3039, opened 2023-01-19
Enable yjit


just1602:
Before we merge #3038, we should add one of the following environment variable to production : `RUBY_YJIT_ENABLE=1` or `RUBYOPT="--yjit"`, so ruby is gonna run with `yjit` enabled, which is supposed to make our application faster and reduce some resources' consumption over time.

veganstraightedge:
✅ I have set `RUBY_YJIT_ENABLE=1 ` in both staging and production 

Thanks @just1602!

veganstraightedge:
I've reverted this change because we're now seeing too much memory usage in production. 

I'll leave this issue open to re-investigate in the future.

<img width="1227" alt="image" src="https://user-images.githubusercontent.com/4361/230801786-9ca3bd02-dffe-4f15-9567-003a2b8d2527.png">

just1602:
What are our options here? Can we increase our manage the quantity of memory every dyno have?

This is possible, I think the JIT compiler have a bigger memory footprint.

veganstraightedge:
I have re-enabled YJIT in staging and production

veganstraightedge:
> What are our options here? Can we increase our manage the quantity of memory every dyno have?
> 
> This is possible, I think the JIT compiler have a bigger memory footprint.

The `WEB_CONCURRENCY` ENV var was set to `4` in production.
For now, I've set down to `3`

bensheldon:
I noticed you're using the jemalloc build pack. Have you set a `MALLOC_CONF`?

You likely want to set:

`MALLOC_CONF="dirty_decay_ms:1000,narenas:2,background_thread:true"`

https://gist.github.com/jjb/9ff0d3f622c8bbe904fe7a82e35152fc

veganstraightedge:
thank you @bensheldon! i just set that in production.

PR #3539, opened 2023-09-14
Fix env var name typo, missing S

veganstraightedge:
Oops.

`RAILS_MAX_THREAD` in puma.rb `RAILS_MAX_THREADS` in env var. 😂


just1602:
@veganstraightedge I'm wondering what were the defaults? Also, would it help to enable `yjit` with a specified number of threads? #3039

PR #3636, opened 2023-10-25
Fixes support subscription cancelation

eliduke:
The stripe gem deprecated subscription `delete`. The new way is now `cancel`.

This should fix the issue with canceling subscriptions, like this...

Screenshot 2023-10-25 at 10 37 12 AM

Yes, I will try canceling my subscription once this goes live.

### DIFF
--- a/app/controllers/support_controller.rb
+++ b/app/controllers/support_controller.rb
@@ -95,7 +95,7 @@ class SupportController < ApplicationController
   def cancel_subscription
     subscription = Stripe::Subscription.retrieve(params[:subscription_id])
 
-    if subscription&.delete
+    if subscription&.cancel
       SupportSession.find_by(token: params[:token]).destroy
       flash.now[:notice] = t('views.support.cancel_subscription.notice')
     else


just1602 (approved):

just1602:
@eliduke I merged it, but I don't have the access to promote to production, so @veganstraightedge will have to do it.

eliduke:
btw, I finally got around to testing this and confirmed that it is indeed working! I was able to cancel my subscription. The only problem now is that I'm realizing that I'm not seeing any `flash` notices appear at any point. We've got a bunch of them in place in the support controller for a number of things. Upon further inspection, I'm seeing a bunch of different syntax like...

```
flash[:notice]
flash.now[:notice]
flash.now.alert
```

Are flash notices still working? Did the syntax get updated? I'm seeing some stuff regarding "themes" (2017 or 2020) and it seems like the flash notice partial isn't available in 2020.

Thoughts?

just1602:
@eliduke `rubocop` enforce `flash.new[:flash_type]` syntax, but both are supposed to work.

Otherwise, you should check in the `2017` theme, the `2020` theme never got fully implemented due to covid and people having less time for the project.

bensheldon:
hopefully helpful: ☺️ 

`flash.now` should be used when the message is to be shown when the current request's template is rendered i.e. it's not a redirect. 

I think that `flash[:type]` likely shouldn't be used, and flash messages instead should be added to the `redirect_to the_path, flash: { type: "message" }`

(And yes, the theme should be double checked that they are rendered)

PR #3704, opened 2023-11-21
Add Dependabot grouped updates

bensheldon:

👍🏻 

I noticed that there are a significant number of Dependabot updates on this repo, more than any other project I contribute to. I think this is a result of doing `daily` updates and `dependency-type: "all"`. (I noticed overwhelm mentioned in https://github.com/crimethinc/website/pull/3511#issue-1884302431)

I wanted to propose a way to make them slightly less onerous (I figured it was easier to just propose several changes directly rather than open an issue, and make any changes to this PR):

* Changes to `interval: weekly`. I usually set this for my projects. This schedule does not apply to Dependabot _security_ updates; they are opened immediately ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval)): 
   > `schedule` defines when Dependabot attempts a new update. However, it's not the only time you may receive pull requests. Updates can be triggered based on changes to your `dependabot.yml` file, or Dependabot security updates. 
* Uses newish ["Grouped update"](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) feature. I split them out into groups of production, development, and lint.

Insights > Dependency Graph > Dependabot. Click on "Recent update jobs" then click on "Check for updates" button.

# Acceptance Criteria

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,9 +9,18 @@ updates:
       # Allow both direct and indirect updates for all packages
       - dependency-type: "all"
     groups:
-      bundler-deps:
-        patterns:
-          - "*"
+      bundler-prod:
+        dependency-type: "production"
+        update-types: minor
         exclude-patterns:
           - rails
-          - stripe
+      bundler-dev:
+        dependency-type: "development"
+        exclude-patterns:
+          - erb_lint
+          - "rubocop*"
+      bundler-lint:
+        patterns:
+          - erb_lint
+          - "rubocop*"        
+        
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,15 @@ updates:
   - package-ecosystem: bundler
     directory: "/"
     schedule:
-      interval: daily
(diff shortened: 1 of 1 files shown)


just1602 (approved):
Oh! This is really cool! :black_heart: 

My only question is : does the `exclude-patterns` on line 15-16 of the `dependabot.yml` file configure dependabot to not open PR for rails update? If it's the case, I'll add rails releases to my rss feed, so'ill know when to update this app. :smile:

bensheldon:
> does the exclude-patterns on line 15-16 of the dependabot.yml file configure dependabot to not open PR for rails update? 

I'm 80% confident that Dependabot will still open non-grouped PRs for dependencies excluded from a particular group.

just1602:
Nice, this is even better if it's the case !

veganstraightedge:
@bensheldon Thanks for jumping in and helping out! 🎉

PR #3748, opened 2023-12-27
Fix labeler config

just1602:


just1602:
Hey @bensheldon, if you happend to have a second, I'd take an hand here.

I don't know if you've already use the labeler action, but I try to update the config to fix the breaking change of the new major version, but the action isn't loading the config from the repo, I think it fetchs the version on the `main` branch  using the Github API.

I try to use the `checkout` action, but it still doesn't work. If you have any idea, let me know, otherwise, I'll come back to that later when I got an idea on how to fix that. :smile:

bensheldon:
@just1602 thanks for the ping 👋🏻 

I just read these parts of the docs which suggest using `on: pull_request` to workaround it, but not my favorite kind of workflow 😓 : https://github.com/actions/labeler#updating-major-version-of-the-labeler

No objections from me to break this out across a couple of pull requests.

just1602:
Thanks @bensheldon , I'll try with the `pull_request` target to make it works, and then rebase to put the right one before trying to merge.

PR #3751, opened 2023-12-30
Reorder Dependabot groups, match minor _and_ patch

bensheldon:

Spongebob's Patrick taking it from over here and pushing it somewhere else

Reorders the Dependabot grouped updates so that the more specific are listed first, followed by the more and more generic groups. This also specifies both minor _and patch_ versions so they properly roll into the group.

I think I now fully understand how Dependabot grouped updates are supposed to work 😓 

# Acceptance Criteria

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,26 +1,26 @@
 version: 2
 updates:
   - package-ecosystem: bundler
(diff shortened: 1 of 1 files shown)


just1602 (approved):
Nice, let's merge this! :)

PR #3799, opened 2024-03-06
TEMP: Rails 7.1 default: add_autoload_paths_to_load_path = true (temporarily)

veganstraightedge:
# Description

Finished framework defaults 7.1 file

# Changes

Overrides a 7.1 default for `config.add_autoload_paths_to_load_path` of `false` with `true`, for now

It needs to be `true` until we rethink how to handle allowing nested directories in locales folders.

But doing this first allows up to bump `load_defaults 7.0` to `7.1`

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -31,6 +31,9 @@ module Crimethinc
     config.load_defaults 7.0
     config.active_support.cache_format_version = 7.1
 
+    # TEMP: delete after load_defaults is 7.1
+    config.add_autoload_paths_to_load_path = false
+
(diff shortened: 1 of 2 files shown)


just1602 (approved):

bensheldon:
Nice find! I'm surprised that the configuration for autoload-paths (for ruby files) would intersect with I18n's paths (for yml files). Glad it works!

veganstraightedge:
@bensheldon I was surprised by that too! 

This is probably good for us anyhow. I've wanted to revisit our locales and yaml files approach, with lessons learned over the past 7 years.

PR #3994, opened 2024-10-22
Add PeerTube URL to Video

veganstraightedge:
- added `peer_tube_url` to `Video` model
- added `video_url` to `Video` model to switch between PeerTube and Vimeo
- added `video_embed_tag` helper
- used `video_embed_tag` in `/videos`, `/videos#show`, `/admin/videos#show`

# TODO

- add a setting in `/admin` be able to switch from PeerTube as the default back to Vimeo, in case of an outage or shutdown

### DIFF
--- a/db/migrate/20241021223548_add_peer_tube_url_to_videos.rb
+++ b/db/migrate/20241021223548_add_peer_tube_url_to_videos.rb
@@ -1,0 +1,5 @@
+class AddPeerTubeUrlToVideos < ActiveRecord::Migration[7.2]
+  def change
+    add_column :videos, :peer_tube_url, :text
+  end
+end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -32,6 +32,7 @@ Rails/OutputSafety:
     - 'app/helpers/articles_helper.rb'
     - 'app/helpers/embeds_helper.rb'
     - 'app/helpers/markdown_helper.rb'
+    - 'app/helpers/videos_helper.rb'
     - 'app/models/article.rb'
 
 # Offense count: 8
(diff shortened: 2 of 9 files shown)


db/migrate/20241021223548_add_peer_tube_url_to_videos.rb
```diff
@@ -0,0 +1,5 @@
+class AddPeerTubeUrlToVideos < ActiveRecord::Migration[7.2]
+  def change
+    add_column :videos, :peer_tube_url, :text
```
just1602, line 3: Isn't text really huge compared to string or they are the same on the database?

just1602:
Also I'm wondering if peertube support on markdown-media is would help, I open an issue years ago about it.

db/migrate/20241021223548_add_peer_tube_url_to_videos.rb
veganstraightedge: it doesn't much matter these days. `text` also gives more flexibility for possibly looooooong URLs. which probably isn't the case with a peertube URL, but it's just a pattern that i always follow for URL columns in a (Postgres) database. 🤙🏻

veganstraightedge:
> Also I'm wondering if peertube support on markdown-media is would help, I open an issue years ago about it.

yes! and in fact, you already made the upstream PR for a specific PeerTube server, `kolectiva.media`
https://github.com/veganstraightedge/markdown_media/pull/19
https://github.com/veganstraightedge/markdown_media/issues/14
🎉

you're right that it would be a good idea to implement a more general PeerTube handler
https://github.com/veganstraightedge/markdown_media/issues/11

when i tested this column's value inside of `/videos` page (in my development environment), it Just Worked™! because of your previous `markdown_media` work! 🎉 🎉 🎉

just1602:
Damn, I didn't even remember I did those PRs. This is great ! :joy:

PR #4043, opened 2024-11-08
Remove automerge CI workflow

veganstraightedge:
I think it's better for now if we are aware of every PR, even/especially little patch updates to dependencies


just1602:
That's good for me, anyways it never really works.

PR #4060, opened 2024-11-14
Dynamically create sitemap.xml

### LINKED ISSUE #1192: Figure out process for updating/uploading the sitemap.xml.gz
Prior art:

- https://github.com/crimethinc/website/issues/123
- https://github.com/crimethinc/website/issues/925
- https://github.com/crimethinc/website/issues/502

`https://crimethinc.com/sitemap.xml.gz` 404s

I'm not sure if we ever actually had the sitemap up on the production site. We might've been waiting until we could figure out a way to not manually run a script and/or have to commit the .xml.gz to source control.

It would be amazing if someone could figure out how to do this at all and even better would be how to automate it with Heroku et al.

### LINKED ISSUE #3780: Sitemap ping failing
Seen in output from:

```
heroku releases:output --remote heroku
```

```
Pinging with URL 'https://crimethinc.com/sitemap.xml.gz':
Ping failed for Google: #<OpenURI::HTTPError: 404 Sitemaps ping is deprecated. See https://developers.google.com/search/blog/2023/06/sitemaps-lastmod-ping.> (URL http://www.google.com/webmasters/tools/ping?sitemap=https%3A%2F%2Fcrimethinc.com%2Fsitemap.xml.gz)
```

veganstraightedge:
# Issues

- https://github.com/crimethinc/website/issues/3780
- https://github.com/crimethinc/website/issues/1192

closes #3780 
closes #1192 

# An investigation and discoveries

So, it turns out that we maybe never had a live sitemap.xml or .xml.gz in production, this whole time. 🤦🏻

The reason that it seemed to work in development and seemed to succeed in the deploy/release in production but not actually available in production is because… Heroku's ephemeral filesystem.

> [Ephemeral Disk](https://devcenter.heroku.com/articles/active-storage-on-heroku#ephemeral-disk)
> Heroku has an “ephemeral” hard drive, this means that you can write files to disk, but those files will not persist after the application is restarted.

So, what was happening was in development, running `bundle exec rails sitemap:create` or `sitemap:refresh` etc, would create the `sitemap.xml.gz` in our local `/public` folders. And stay there. Seems good.

In production, during the release stage of a production deploy (as defined in the [`Procfile`](https://github.com/crimethinc/website/blob/5c88026595d2ad22ea872a8427063ff617e851fe/Procfile#L1)), the `sitemap:refresh` would "succeed", but then the `/public` folder it was created in wouldn't necessarily be in the actual dyno/s serving any real requests. AFAICT.

---

# My preferred requirements

When working on this, I went round and round trying to make it work with all of these conditions:

- [sitemap_generator](https://github.com/kjvarga/sitemap_generator) gem (great gem, love it, would use in non-Heroku deployed apps)
- Heroku's ephemeral filesystem
- and keeping its location on the same domain at the root, where it can be auto-discovered without first reading `robots.txt`

The gem [suggests and has functionality](https://github.com/kjvarga/sitemap_generator?tab=readme-ov-file#upload-sitemaps-to-a-remote-host-using-adapters) to store the generated file somewhere else (say, S3), but I'd like to keep it in its _well known_ location. 

---

# Conclusion

In the end, I decided to create a sitemaps controller and dynamically create the file to:

- keep it at `/sitemap.xml`
- not have to deal with Heroku's ephemeral filesystem
- keep its discovery well known and not dependent on `robots.txt` (a separate issue/pr to do!)

The challenge and risk, of course, is performance. Namely around articles and some of the tools (zines, etc) which have the biggest tables to scan. Especially since most items in the sitemap never change. 

But that _not ever really changing-ness_ is what allowed me to use Rails' fragment caching around each `<url>` item in the long list of `<urlset>` and reduce the page load time from ~1s to ~200ms (depending on warm cache, etc). Even at 1s, it's not the end of the world, since (I'm suspecting) that this file doesn't get read a ton.

# TODO follow up

- [ ] remove `sitemap:refresh` from `Procfile`
- [ ] remove `config/sitemap.rb`
- [ ] remove `sitemap_generator` gem
- [ ] keep an eye on this in production
- [ ] add any files/paths that i missed (or skipped for now because i ran out of steam)


just1602 (approved):

veganstraightedge:
✅ https://crimethinc.com/sitemap.xml

PR #4191, opened 2024-12-27
Add missing domains and subdomains to redirection middleware

veganstraightedge:


app/middlewares/rack/domain_redirect.rb
```diff
@@ -1,42 +1,60 @@
...
+      # ccTLDs => localized subdomain
+      ['es.', '', /crimethinc\.es/],
+      ['de.', '', /crimethinc\.de/],
+      ['cs.', '', /cz\.crimethinc\.com/], # Fix our orignal mistaken assumption
+
+      # short domain (for historical twitter/etc posts)
+      ['', '', /cwc\.im/]
+    ].freeze
+
+    CC_TLD_SUBDOMAINS = %w[
```
veganstraightedge, line 28: It'd be better if we could read this a file/value in the repo
And used for both this _and_ for allowed locales

Issue #1392, opened 2019-12-14
Convert TCE PDF to web version: Malay / Malay


veganstraightedge:
Malay / Malay
http://www.crimethinc.com/blog/2016/09/21/to-change-everything-in-11-more-languages/#malay

See for details:
https://github.com/crimethinc/website/issues/575

Issue #575, opened 2017-12-14
Create web versions for all TCE editions


astronaut-wannabe:
## languages 
- [ ] العربية
- [ ] فارسی
- [ ] հայերեն

- [ ] Български
- [ ] Cebuano
- [ ] Català
- [ ] 中文
- [ ] Dansk
- [ ] Español (América Latina)
- [ ] Français
- [ ] ελληνικά
- [ ] Italiano
- [ ] 한국어
- [ ] Malay
- [ ] Malti
- [ ] Nederlands
- [ ] Română
- [ ] русский
- [ ] Srpskohrvatski
- [ ] Slovenčina
- [ ] Svenska
- [ ] Tagalog

astronaut-wannabe:
PR #577 will add the infrastructure to do this

just1602:
Have we a plan to handle language duplication like Français vs Français (Québec) and Espanol (America Latina) vs Espanol (spanish) or we just put one of them and not both?

astronaut-wannabe:
> Have we a plan to handle language duplication like Français vs Français (Québec) and Espanol (America Latina) vs Espanol (spanish) or we just put one of them and not both?

@just1602  we will do both. I know for sure this will work:

within `es.yml`

```yaml
espanol: # this is whatever the url is going to be
  layouts:
    to_change_everything:
      language_direction: ltr
      lang_code: es
# etc. etc.
espanol-america-latina:
  layouts:
    to_change_everything:
      language_direction: ltr
      lang_code: es
# etc. etc.
```

the only thing I am not sure is if I will need to have a full set of the yaml for dialects, or if I can, for example, use espanol as the base and only override what differs.

but either way, the current plan is to do unique pages for unique dialects/regions/etc

veganstraightedge:
When this is done, close https://github.com/crimethinc/website/issues/548

veganstraightedge:
@Bargraph6 @just1602 I re-organized list of TCE languages above. 
And I created sever draft PRs with the beginning of the YAML file with text extracted as best as I could automated, from .docx, .doc, .odt, and .indd/.idml files. 

They definitely still need to be manually reviewed and completed. But hopefully it will help someone get a jumpstart on each languages.

PR #4758, opened 2025-08-04
Create web version of the Malaysian TCE pdf

Bargraph6:
## This change

Malay comes in many dialects, so I am *assuming* "standard malay" is used in the pdf.

The generator and guide appears to be working well, though the
`video_url` in the generated yaml is in the wrong section. The
generator should be updated to correct that.

## Resources used

- https://github.com/crimethinc/website/blob/main/docs/to-change-everything-guide.md
- https://en.wikipedia.org/wiki/Malaysian_Malay

relates to: https://github.com/crimethinc/website/issues/575
closes: https://github.com/crimethinc/website/issues/1392


just1602:
It looks good to me, but I'll let @veganstraightedge review it since he did way more of those then me.

veganstraightedge:
> using this PR as an example, I went with bahasa-melayu-piawai, but perhaps that is too formal and malay would be fine? On the other hand, I don't want to just use malay if the translation is in a specific dialect..

heard. could we change this to just `Malay` to begin with. if a more specific dialect name is better suited, then we can update/redirect in the future. :)

other than that, this looks great and will be ready to merge/ship!

veganstraightedge:
Great job! 🎉 Thank you so much for doing this.

Amazing that you're already starting another one too #4805 ⭐️

veganstraightedge (approved):

PR #4764, opened 2025-08-05
Remove unused references from the asset manifest file

Bargraph6:
## Context

When I initially set up a development environment, I got an [`execjs`][0] error telling me to install a JS runtime. I did that, and it made me curious as to what the runtime is needed for

While looking at the asset manifest, I noticed some no-longer-used references in there.

## This change

1. I removed the lite_mode reference per the TODO. When testing locally the styling still appears to work fine.
2. I removed the `vendor` directory as it is unused

### How I tested the `lite` domain[^1]

<table>
  <thead><tr><th>demo screen recording</th></tr></thead>
  <tbody>
    <tr><td>

https://github.com/user-attachments/assets/30eae829-6ea2-4cf1-9ee7-ac2826c113ee

</td>
</tr>
  </tbody>
</table>

I tested locally by modifying my `/etc/hosts` to use a local domain:

```
127.0.0.1	localhost localhost.localdomain lite.localhost.localdomain
::1		localhost localhost.localdomain lite.localhost.localdomain
```

I then added the domain to `config.hosts` in
`config/environments/development.rb`:

```ruby
  # use subdomains with localhost (e.g. http://lite.localhost:3000)
  config.hosts << "localhost.localdomain:3000"
  config.hosts << /.*\.localhost\.localdomain:3000/
```

[^1]: Setting `config.action_dispatch.tld_length = 0` is an
      easier/quicker work-around to use subdomains on localhost. This
      doesn't require modifying `/etc/hosts` or adding additional values to
      `config.hosts`. However, it will (probably) break the `link_for` helper
      used in the router in non-obvious ways

[0]:https://github.com/rails/execjs

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -14,7 +14,5 @@
 // 2025 theme
 //= link_tree ../stylesheets/2025 .css
 
-// TODO: delete this?
-// ??? lite_mode.css
+// javascript used for support page and live blogs
 //= link_tree ../../javascript .js
-//= link_tree ../../../vendor/javascript .js


veganstraightedge:
thanks @Bargraph6! 🎉

PR #5066, opened 2026-01-08
ignore nix stuff

just1602:
I switched from git to jj for this project, and since there's no staging area in jj and everything is staged and add unless it's actively ignored, this commit ignore nix stuff since I don't want to pollute the repo with it for now.


veganstraightedge:
thanks for adding these!

PR #5105, opened 2026-01-29
Rubocop: Rails/EnvironmentVariableAccess -> config files

veganstraightedge:
Rubocop: Rails/EnvironmentVariableAccess: https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsenvironmentvariableaccess

- added a yaml config loader to `application.rb` using  `config_for` 
- added a yaml file per every "service" provider, with section for each environment: dev, test, prod
- replaced all `ENV[]` or `ENV.fetch` with `Rails.application.config.x.[NAMESPACE].[KEY]`

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -112,21 +112,12 @@ module Crimethinc
     # To opt in to the new behavior, set `config.active_support.to_time_preserves_timezone = :zone`.
     config.active_support.to_time_preserves_timezone = :zone
 
-    # Load each service specific config file into the app’s config,
-    # using Rails’ custom config namespace: x
+    # Load service-specific config files into Rails' custom config namespace: x
     # Example: Rails.application.config.x.stripe.secret_key
-    #
-    # Enumerate through all YAML files in config/services
-    Rails.root.glob('config/services/*.yml').each do |yaml_filepath|
-      # Read the service name from the YAML file's name
-      service_name = File.basename(yaml_filepath, '.yml')
-
-      # Read the YAML config file for this service
-      service_config = config_for("services/#{service_name}")
-
-      # Load the service's configuration into the custom config namespace
-      # Example: when `yaml_filepath` is "stripe.yml", this will call `config.x.stripe =`
-      config.x.public_send(:"#{service_name}=", service_config)
-    end
+    config.x.app         = config_for 'services/app'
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -111,5 +111,22 @@ class Application < Rails::Application
...
+    Rails.root.glob('config/services/*.yml').each do |yaml_filepath|
+      # Read the service name from the YAML file's name
+      service_name = File.basename(yaml_filepath, '.yml')
+
+      # Read the YAML config file for this service
+      service_config = config_for("services/#{service_name}")
+
+      # Load the service's configuration into the custom config namespace
+      # Example: when `yaml_filepath` is "stripe.yml", this will call `config.x.stripe =`
+      config.x.public_send(:"#{service_name}=", service_config)
```
just1602: Is the `x` needed? I'm just curious if it would be possible to simply do `config.public_send` for example and be able to do `config.stripe =` for example?

just1602 (dismissed):

config/application.rb
veganstraightedge: The `config.x` namespace is the recommendation in the Rails guides. If there's nesting in the config. So far there's not. But the other thing that I like about `.x.` is that avoids possible namespace collision with Rails configs. 

https://guides.rubyonrails.org/configuring.html#config-x

https://guides.rubyonrails.org/configuring.html#custom-configuration

veganstraightedge:
@just1602 good point! I un-metaprogrammed it. Since we only have a few config files.

config/application.rb
```diff
@@ -111,5 +111,13 @@ class Application < Rails::Application
...
     config.active_support.to_time_preserves_timezone = :zone
+
+    # Load service-specific config files into Rails' custom config namespace: x
+    # Example: Rails.application.config.x.stripe.secret_key
+    config.x.app         = config_for 'services/app'
+    config.x.bugsnag     = config_for 'services/bugsnag'
+    config.x.rack_attack = config_for 'services/rack_attack'
+    config.x.redis       = config_for 'services/redis'
+    config.x.stripe      = config_for 'services/stripe'
   end
```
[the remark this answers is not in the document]
veganstraightedge: ah ok. gotcha. 

another available/future use case for these config YAML files: non-secret static values for different environments 

- not in code
- not in ENV vars
-

PR #5216, opened 2026-03-25
css refactor: convert border-size sccs vars to css vars

Bargraph6:

related to this PR: https://github.com/crimethinc/website/pull/5212

Most of this is just a find/replace. There is one additional change in the podcasts component where I also converted the calc to avoid mixing scss syntax with the css syntax:

```diff
     .latest-episode {
-      margin: (2 * $border-size-xlarge) auto;
+      margin: calc(var(--border-size-xlarge) * 2) auto;
```

Since this change is converting scss variables to css variables, there are only a few things that can go wrong:

1. The converted file isn't loading anymore
2. the variables are not being applied
3. specifically for the podcast change, the `calc` value is generating a different value in css vs in scss

to check these things you can load the site with this change and:

1. see the new variables in the `:root` element in dev tools
2. toggle the css variables in dev tools to see that they are being applied
3. for the podcast calc value, you can verify the same value is generated manually as show in the below screenshots

| before | after |
|------------|---------|
| before1-compressed | after1-compressed |
| before2-compressed | after2-compressed |

This is the first in what will be an extended series of commits to eventually drop the css and js pre-processor gem dependencies as discussed in https://github.com/crimethinc/website/pull/5212

This first series of PRs will remove the scss specific syntax in-place, ultimately ending in a PR the renames all of the `.scss` files to `.css`

# Acceptance Criteria

### DIFF
--- a/app/assets/stylesheets/2017/base/variables/_borders.scss
+++ b/app/assets/stylesheets/2017/base/variables/_borders.scss
@@ -1,7 +1,7 @@
 :root {
-    --border-size-xsmall:  1px;
-    --border-size-small:   5px;
-    --border-size-medium: 10px;
-    --border-size-large:  20px;
-    --border-size-xlarge: 40px;
+  --border-size-xsmall:  1px;
+  --border-size-small:   5px;
+  --border-size-medium: 10px;
+  --border-size-large:  20px;
+  --border-size-xlarge: 40px;
 }


Bargraph6:
@veganstraightedge @just1602 here is the first in a series to convert scss to css.

the borders and the colors (which will be the next PR) are the changes that seem to touch the most files, however the changes are almost entirely a find/replace so pretty low risk.

starting with borders and colors so that subsequent PRs have much less noise in the diffs

just1602 (approved):

PR #5280, opened 2026-04-20
css refactor: Rename several SCSS files to .css

Bargraph6:
What does this commit do
===========================

Since the [conversion of the social icons from SCSS to CSS][0] has reached production and is working, we can move forward with renaming several `.scss` files to `.css`.

**This is only a rename**.  No changes are needed for these files as they already contains only CSS markup.

There are other files that contain only CSS that need to be renamed. However, since they are larger files I will rename them in individual commits in case there is some SCSS syntax I missed.

What are the relevant GitHub issues?
===========================

related to: https://github.com/crimethinc/website/pull/5212
related to: https://github.com/crimethinc/website/pull/5259
related to: https://github.com/crimethinc/website/pull/5223
related to: https://github.com/crimethinc/website/pull/5216

Is there any background context you want to provide for reviewers?
===========================

### Changelog

* `2017/articles_custom_css.css`: This file is rather small, so it easy to see that it only contains CSS. I also tested this locally and it still works as a CSS file.
* `2017/base/variables/_borders.scss`: previously converted to css in [ba7c239][1].  Doing rename now as the social icons test worked in production.
* `2017/base/variables/_colors.scss`: previously converted to css in [eae7461][2].  Doing rename now as the social icons test worked in production.
* `2017/components/_buttons.scss`: With the previous commits to replace border and color SCSS variables with CSS variables done, this file now only contains CSS and can be renamed.
* `2017/components/_colors.scss`: This file is rather small, so it is easy to see that it only contains CSS.
* `2017/components/_localization.scss`: This file is very small (3 lines), so it is easy to see that it only contains CSS.
* `2017/lib/_reset.scss`: As referenced in the header comment, this file is a copy/paste of a CSS file.
* `2017/lib/_utilities.scss`: This file is very small (3 lines), so it is easy to see that it only contains CSS.

# Acceptance Criteria
## These should be checked by the reviewers

- [ ] This pull request does not cause the database export script to become out of sync with the db schema

[0]:https://github.com/crimethinc/website/pull/5259 
[1]:https://github.com/crimethinc/website/commit/ba7c2397a77a48f11a1792e5885bbf7885ea5a66 
[2]:https://github.com/crimethinc/website/commit/eae74618f7658be221e7da2f69e232c573aae724

PR #5259, opened 2026-04-14
css refactor: components/_social

Bargraph6:
# What are the relevant GitHub issues?

related to this PR: https://github.com/crimethinc/website/pull/5212

# What does this pull request do?

This PR:

1. renames `_socials.scss` to `_socials.css`
2. updates the social icons to use css variables instead of scss variables
3. adds a `.visually-hidden` css class to use in place of the current `text-indent: -10000%;` approach

# How should this be manually tested?

I loaded up the home page and articles page to see the social links:

| | before | after |
|-|--------|-------|
| **home** | before-home | after-home |
| **article** | before-article | after-article |

# Questions for the pull request author (delete any that don't apply):
- [ ] Are any needed README/wiki/documentation updates needed for this pull request?
- [ ] Does the code you updated have tests? If not, could you add some please?
- [ ] Does this pull request require any new server dependencies which need to be added to the build process? If so, please explain what.

# Acceptance Criteria
## These should be checked by the reviewers

- [x] This pull request does not cause the database export script to become out of sync with the db schema

PR #5290, opened 2026-04-23
css refactor: Convert _video.scss to _video.css

Bargraph6:
===========================

- related to: https://github.com/crimethinc/website/pull/5212
- related to: https://github.com/crimethinc/website/pull/5280

===========================

Since the [conversion of the social icons from SCSS to CSS][0] has reached production and is working, we can move forward with converting and renaming the remaining `.scss` files to `.css`.

This change converts the `_video.scss` file to CSS by :

1. unwrapping the `font-*` styles to CSS.

===========================

### Changelog

* 2017/_video.css: Converted the SCSS `font {...}` syntax to CSS.

[0]:https://github.com/crimethinc/website/pull/5259

### DIFF
--- a/app/assets/stylesheets/2017/views/_video.css
+++ b/app/assets/stylesheets/2017/views/_video.css
@@ -1,10 +1,8 @@
 /* TODO: Is this used anywhere? */
 #video {
   h1 {
-    font: {
-      size: 1.25em;
-      weight: 900;
-    }
+    font-size: 1.25em;
+    font-weight: 900;
     letter-spacing: -.01em;
     line-height: 1.2em;
     margin-bottom: .1em;
@@ -12,8 +10,8 @@
 }
 
 .video-flexbox-container {
-    display: flex;
-    flex-direction: row;
-    flex-wrap: wrap;
-    justify-content: space-between;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: space-between;
 }


app/assets/stylesheets/2017/views/_video.css
```diff
@@ -1,19 +1,17 @@
 /* TODO: Is this used anywhere? */
```
Bargraph6, line 1: I tried to investigate this by running the following code in the rails console:

```ruby
matches = [] 
Article.all.find_each do |a|
  content = a.content_rendered(include_media: true)
  search = /.*id="?(\w*)"?/i
  matches << "id=#{content.match(search)[1]}" if content.match?(search)
end
matches.uniq
```

I didn't see any usages of `id="video"`, however:

1. I am only searching articles
2. this is only searching `article.content`
  - there are other resource types (e.g video, tools)
  - there is a lot of html surrounding of `.content` (the erb views/partials/templates), which I also tried searching but couldn't be sure I covered everything.
3. I was only searching against my local db dumb, which is not the full production DB

Bargraph6:
@just1602 here is a small one ready for review

app/assets/stylesheets/2017/views/_video.css
just1602: I'd say, for now we can merge the conversion, and open an issue to make sure this is used or remove it while not blocking the progress of this migration.

But a good way to know might be to remove the file and test locally that all videos are displayed correctly.

just1602 (approved):

veganstraightedge:
`id=video` is on `body` for videos#show page:

```html
<body id="video">
```

https://crimethinc.com/videos/mercenaries

PR #5295, opened 2026-04-24
2025 layout: handle admin header for non-localizable models

Bargraph6:
=======================

n/a

====================

* `app/views/layouts/2025/_admin_header.html.erb`: skip rendering the localization links in the admin header if the active record model doesn't define `.localizations`

1. **without patch**
  - be logged in as a publisher
  - set theme to 2025
  - go to http://localhost:3000/categories
  - see error
2. **with patch applied**
  - be logged in as a publisher
  - set theme to 2025
  - go to http://localhost:3000/categories
  - see page render, admin banner should have no localization options

While working on pagination CSS changes related to the work to drop the sass preprocessor dependency[^1], I ran into a `500` when trying to view the categories index.

In this scenario, the `@editable` is set to a `Category` model, which does not (currently) `include Translatable`.

Acceptance Criteria
=======================

[^1]:https://github.com/crimethinc/website/pull/5212

### DIFF
--- a/app/views/layouts/2025/_admin_header.html.erb
+++ b/app/views/layouts/2025/_admin_header.html.erb
@@ -13,7 +13,7 @@
     </div>
 
     <div>
-      <% if @editable.present? %>
+      <% if @editable.present? && @editable.respond_to?(:localizations) %>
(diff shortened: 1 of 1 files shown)


Bargraph6:

just1602 (approved):

PR #5298, opened 2026-04-26
pagination: improve pagination semantics for screen readers

Bargraph6:
=============================================

- related to #5212 

==================================

### first commit | pagination: give pagination semantics for screen readers

* `app/views/2017/articles/show.html.erb`: change pagination container
from a div to nav element. Also add an aria-label.
* `app/views/2017/home/index.html.erb`: change pagination container
from a div to nav element. Also add an aria-label.
* `app/views/2025/articles/show.html.erb`: change pagination container
from a div to nav element. Also add an aria-label.
* `app/views/kaminari/_paginator.html.erb`: change pagination container
from a div to nav element. Also add an aria-label
* `config/locales/en/en.yml`: add english translation of new
aria-label

By changing the `<div>` to a `<nav>` for the pagination container, screen
readers have more of what this is. I also added an optional label.

Rather than let the label fall back to english, I provided a blank
string as the default. This makes the aria-label behave like the
browser default (i.e. it is just skipped), rather than having english
read out when the language is something else.

With a screen reader the change will be read something like:

- "end navigation" when there is no translation for the aria-label
- "end navigation pagination" when there is a translation

### second commit:

pagination: copy aria enhancements from twitter-bootstrap-4 to 2017

* `app/views/kaminari/_next_page.html.erb`: for screen readers, don't
read out the literal icon, rather read out "previous page"
* `app/views/kaminari/_prev_page.html.erb`: for screen readers, don't
read out the literal icon, rather read out "next page"

Prior to this change, screen readers would read out something like
"right facing caret ...". With the change it now reads out something
like "Next ...".

These changes already exist in
`app/views/kaminari/twitter-bootstrap-4/_next_page.html.erb` and
`app/views/kaminari/twitter-bootstrap-4/_prev_page.html.erb`, so I
just copied them up to the view partial being used on the 2017 theme.

How should this be manually tested?
=============================================

- If you don't already have a screenreader set up, you can turn on the screen reader built into most smartphone operating systems
- you can inspect the accessibility tree in the browser dev tools

### screen recording (includes sound)

| before | after |
|-------------|---------|
| https://github.com/user-attachments/assets/a32046ae-483a-4883-9008-4fc0757edf34 | https://github.com/user-attachments/assets/0e16dc7c-3687-4c68-883f-bca28d701b52 | 

### DIFF
--- a/app/views/kaminari/_next_page.html.erb
+++ b/app/views/kaminari/_next_page.html.erb
@@ -6,5 +6,10 @@
     per_page:      number of items to fetch per page
     remote:        data-remote -%>
 <li class="next">
-  <%= link_to_unless current_page.last?, t("views.pagination.next").html_safe, url, rel: "next" %>
+    <% unless current_page.last? %>
+      <%= link_to url, rel: "next", class: "page-link" do %>
+        <span aria-hidden="true"><%= t("views.pagination.next").html_safe %></span>
+        <span class="visually-hidden"><%= t("views.pagination.next_screen_reader").html_safe %></span>
+      <% end %>
+  <% end %>
 </li>
--- a/app/views/kaminari/_paginator.html.erb
+++ b/app/views/kaminari/_paginator.html.erb
@@ -5,8 +5,11 @@
     per_page:      number of items to fetch per page
     remote:        data-remote
     paginator:     the paginator that renders the pagination tags inside -%>
+
+<% container_label = t('views.pagination.label', default: '') %>
+
 <%= paginator.render do -%>
-  <div class="pagination-container">
+  <nav class="pagination-container" aria-label="<%= container_label %>">
     <ul class="pagination">
       <%= prev_page_tag unless current_page.first? %>
       <% each_page do |page| -%>
@@ -20,5 +23,5 @@
         <%= next_page_tag unless current_page.last? %>
       <% end %>
     </ul>
-  </div>
+  </nav>
 <% end %>
--- a/app/views/kaminari/_prev_page.html.erb
+++ b/app/views/kaminari/_prev_page.html.erb
@@ -6,5 +6,10 @@
     per_page:      number of items to fetch per page
     remote:        data-remote -%>
 <li class="prev">
-  <%= link_to_unless current_page.first?, t("views.pagination.previous").html_safe, url, rel: "prev" %>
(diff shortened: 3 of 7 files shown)


app/views/kaminari/_paginator.html.erb
```diff
@@ -5,8 +5,11 @@
     per_page:      number of items to fetch per page
     remote:        data-remote
     paginator:     the paginator that renders the pagination tags inside -%>
+
+<% container_label = t('views.pagination.label', default: '') %>
```
Bargraph6, line 9: **note:** The call to `t(...)` must be made outside of the kaminari `paginator.render` block because kaminari defines a `t()` function with a different parameter list.

app/views/kaminari/_next_page.html.erb
```diff
@@ -6,5 +6,10 @@
     per_page:      number of items to fetch per page
     remote:        data-remote -%>
 <li class="next">
-  <%= link_to_unless current_page.last?, t("views.pagination.next").html_safe, url, rel: "next" %>
+    <% unless current_page.last? %>
```
Bargraph6, line 9: **note:** copied from https://github.com/crimethinc/website/tree/578f983e9a8fb3c45e2a6a87e605761a7e258232/app/views/kaminari/twitter-bootstrap-478

app/views/kaminari/_prev_page.html.erb
```diff
@@ -6,5 +6,10 @@
     per_page:      number of items to fetch per page
     remote:        data-remote -%>
 <li class="prev">
-  <%= link_to_unless current_page.first?, t("views.pagination.previous").html_safe, url, rel: "prev" %>
+  <% unless current_page.first? %>
```
Bargraph6, line 9: **note:** copied from https://github.com/crimethinc/website/tree/578f983e9a8fb3c45e2a6a87e605761a7e258232/app/views/kaminari/twitter-bootstrap-478

Bargraph6:
@just1602 @veganstraightedge this is ready for review. while testing I noticed some funkiness around the pagination on the article page, however I have confirmed it is unrelated to this PR and is broken in production. I have a separate PR (#5299) that addresses that issue

just1602 (approved):

Issue #5304, opened 2026-04-27
[bs#681063] Missing config/locales/pages/ca/litkit_main.markdown


veganstraightedge:
# bs#681063

https://app.bugsnag.com/crimethinc/rails/errors/685caa10682d944195681063

`books#lit_kit` — `ActionView::Template::Error: No such file or directory @ rb_sysopen - /app/config/locales/pages/ca/litkit_main.markdown`

5992 occurrences. `ca` (Catalan) locale of the litkit page is referenced but the markdown file doesn't exist. Either add the file or skip this locale path.

## Related grouping (same root cause, different locale)

https://app.bugsnag.com/crimethinc/rails/errors/69effa7dba7b406400c7a773

`pages#about` — `No such file or directory @ rb_sysopen - /app/config/locales/pages/es-419/about.markdown`. 49× occurrences. Same locale-fallback bug — fixed by the same patch in #5311.

veganstraightedge:
Fixed by #5311 (deployed 2026-04-28). Resolves both Bugsnag groupings: `685caa10682d944195681063` (`ca` locale) and `69effa7dba7b406400c7a773` (`es-419` locale).

PR #5311, opened 2026-04-27
[#5304] Fall back to default locale when localized markdown is missing

veganstraightedge:
## Summary

- `MarkdownHelper#render_markdown_for` was reading `config/locales/pages/{I18n.locale}/{page}.markdown` directly with `File.read`, no fallback. Locales without a localized markdown file (e.g. `:ca`, `:el`, `:hy`, `:mt`) 500'd on `books#lit_kit` and any other page using this helper.
- Rails `config.i18n.fallbacks` only applies to `I18n.t` lookups, not raw filesystem reads.
- Fix: check for the localized file's existence; fall back to `I18n.default_locale` if missing.

Refs #5304 (Bugsnag bs#681063 — 5992× occurrences).

## Test plan

- [ ] `bundle exec rspec spec/helpers/markdown_helper_spec.rb`
- [ ] Visit `/ca/books/lit-kit` in staging — should render English content instead of 500
- [ ] Visit `/en/books/lit-kit` — unchanged

### DIFF
--- a/app/helpers/markdown_helper.rb
+++ b/app/helpers/markdown_helper.rb
@@ -1,8 +1,9 @@
 module MarkdownHelper
   def render_markdown_for page:
-    content = File.read [Rails.root, "config/locales/pages/#{I18n.locale}", "#{page}.markdown"].join('/')
+    path = Rails.root.join "config/locales/pages/#{I18n.locale}/#{page}.markdown"
+    path = Rails.root.join "config/locales/pages/#{I18n.default_locale}/#{page}.markdown" unless path.exist?
 
-    render_markdown content
+    render_markdown File.read(path)
   end
 
   def render_markdown text, remove_wrapper_p_tag: false
--- a/app/views/2017/home/index.html.erb
+++ b/app/views/2017/home/index.html.erb
@@ -24,7 +24,7 @@
   <% end %>
 </div>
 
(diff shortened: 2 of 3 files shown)


app/helpers/markdown_helper.rb
```diff
@@ -1,8 +1,9 @@
 module MarkdownHelper
   def render_markdown_for page:
-    content = File.read [Rails.root, "config/locales/pages/#{I18n.locale}", "#{page}.markdown"].join('/')
+    path = Rails.root.join "config/locales/pages/#{I18n.locale}/#{page}.markdown"
+    path = Rails.root.join "config/locales/pages/#{I18n.default_locale}/#{page}.markdown" unless path.exist?
```
just1602, line 4: Is it normal that you define the same path twice.

just1602 (dismissed):

app/helpers/markdown_helper.rb
veganstraightedge: it's really an if/else

L3: set `path`
L4: set `path` to a different thing, if there is no file at `path`

just1602: Sorry, damn, I first reviewed it on my cellphone and didn't see the `unless path.exist?` at the end of the line.

just1602 (approved):

PR #5314, opened 2026-04-29
css refactor: convert _pagination.scss to CSS

Bargraph6:
## What are the relevant GitHub issues?

related to: https://github.com/crimethinc/website/pull/5212

===================================

* `app/assets/stylesheets/2017/components/_pagination.css`: renamed `_pagination.scss` to `_pagination.css`.  Then updated the styles to use flexbox.
* `app/views/kaminari/_page.html.erb`: modified this template to remove the disabled link from keyboard focus, as well as marking disabled to screen readers.
* `config/locales/en/en.yml`: add a i18n string to support the kaminari template change.

This change consists of two major parts. First, modifying the kaminari template, and then converting the pagination SCSS to CSS, accounting for the template changes.

### The kaminari template changes

1. updated the page links for screen readers so that they now read e.g. "page one" as opposed to "one"
2. the link for the current page was not actually a link prior to this change, it was a `<div>`. This was done to make the current page not clickable. It was however still focusable with keyboard navigation, as well as still read out by screen readers. Rather than having a `<div>`, I kept the link as an `<a>`, opting to disable to click actions in CSS[^1]. I then removed it from keyboard focus and marked as disabled to screen readers[^2].

### The CSS change

The main difference with the CSS, besides accounting for the template changes, is to make use of flexbox.

There are some additional changes that can probably be made for article navigation specifically, but those can be in a separate patch.

- check pagination on the home screen at various screen sizes
- check pagination on an index page (e.g. http://localhost:3000/categories/analysis )
- check pagination on an aticle
  - especially one with a long link (e.g. http://localhost:3000/2026/01/21/from-rapid-response-to-revolutionary-social-change-the-potential-of-the-rapid-response-networks )
- this shouldn't have any impact on the 2025 theme pagination, but that should be double checked

The home page pagination and in index page pagination should essentially look the same visually.

The article page is different, as I changed that layout a bit. Here is how the article pagination looks currently in production when one of the links is on the longer side of things:

| current production | production with some visual annotations |
|--------|--------|
| production | production-with-annotations | 

I have changed the layout so that the next/previous buttons take up 50% of the pagination container when they can, but if the screen is small enough to where `50% < 40ch`, then the buttons will wrap to be on top of each other.

| with patch desktop | with patch phone |
|--------|--------|
|  with-patch-desktop |  with-patch-phone | 

I think some further tweaks can be made to the article pagination, such as:

- moving the "< previous" and "next >" to `::before`/`::after` pseudo-elements 
  - and positioning them either above, or on the sides of the link
- styling the "< previous" and "next >" a bit differently to make them stand out

however I opted to defer that for now as the PR was already a good sized diff to review.

---

[^1]: This is because a disabled <a> is more semantic for screen readers. 
[^2]: The `tabindex=-1` attribute is used to remove it from keyboard focus. `aria-disable` indicates the link is disabled to screen readers, while `aria-current` is used to indicate it is a link to the current page.


Bargraph6:

just1602 (approved):
Nice to see you also converted it to flex!

PR #5321, opened 2026-04-30
[#5308] Redirect to podcasts index when slug not found

### LINKED ISSUE #5308: [bs#7ea24d] Podcast slug not found should redirect, not 500
# bs#7ea24d

https://app.bugsnag.com/crimethinc/rails/errors/69c68e036bb8e831667ea24d

`ActiveRecord::RecordNotFound: Couldn't find Podcast with [WHERE "podcasts"."slug" = $1]`. 10 occurrences.

Currently raising as an `info` event but it shouldn't be reaching Bugsnag at all — missing podcast slug should 404 cleanly (rescue `ActiveRecord::RecordNotFound` in the controller) or redirect to the podcast index.

veganstraightedge:
# #5308

https://app.bugsnag.com/crimethinc/rails/errors/69c68e036bb8e831667ea24d

## Summary

- `PodcastsController#show` was using `Podcast.find_by!`, which raised `ActiveRecord::RecordNotFound` when a slug had no matching podcast. Bugsnag captured this as a recurring `info`-level error.
- Switched to `Podcast.find_by` + early redirect to the podcasts index when the slug is unknown. Mirrors the existing pattern in `BooksController#set_book`.
- Updated the podcast factory to set a `slug` (the model validates `presence: true`, so the factory was previously incapable of producing a valid record).
- Added `spec/controllers/podcasts_controller_spec.rb` covering both the happy path and the unknown-slug redirect.

## Test plan

- [ ] In staging, visit `/podcasts/no-such-podcast` — should redirect to `/podcasts` instead of returning 500

### DIFF
--- a/app/controllers/podcasts_controller.rb
+++ b/app/controllers/podcasts_controller.rb
@@ -10,9 +10,11 @@ class PodcastsController < ApplicationController
   end
 
   def show
+    @podcast = Podcast.find_by slug: params[:slug]
+    return redirect_to(podcasts_path) if @podcast.blank?
+
     @html_id  = 'page'
     @body_id  = 'podcast'
-    @podcast  = Podcast.find_by! slug: params[:slug]
     @episodes = @podcast.episodes.live.published
     @title    = PageTitle.new [title_for(:podcasts), @podcast.name]
 
--- a/spec/controllers/podcasts_controller_spec.rb
+++ b/spec/controllers/podcasts_controller_spec.rb
@@ -1,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe PodcastsController do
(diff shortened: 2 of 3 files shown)


just1602:
If we want to do this, I'm not opposed to it, since it would be consistent with the rest of the site. But sometime, I'm wondering if we shouldn't display the 404 page instead, because a couple of times I tried to access some resource directly from URL, and it took me a little moment to understand that I was redirected to the index of the resource because the URL was wrong and not due to a bug in the cms app logic.

EDIT: I'll merge this, because it's better to match the behaviour of the whole app, but it can be worth it to ask ourselves if it's the behaviour we want.

just1602 (approved):

Issue #5303, opened 2026-04-27
Add Rack middleware to drop malformed bot/scraper requests before Rails


veganstraightedge:
Production Bugsnag is dominated by clearly bot/scraper-driven noise that's currently surfacing as Rails errors. Instead of fixing each as it appears, drop these at the Rack layer before they hit the application.

### Malformed requests
- bs#8655d6 — `InvalidType` junk MIME header (`cacheposcontenttypecachepos<>`) — 7253×
- https://app.bugsnag.com/crimethinc/rails/errors/67ac7d93f90d606d6b8655d6
- bs#af2992 — `UnsafeRedirectError` w/ binary garbage in URL — 3069×
- https://app.bugsnag.com/crimethinc/rails/errors/695cd6544ef4ae6af7af2992
- https://app.bugsnag.com/crimethinc/rails/errors/69efececba7b406400c3ddad
- bs#44a9c2 — `ArgumentError` malformed UTF-8 in `string.unpack("U*")` — 1175×
- https://app.bugsnag.com/crimethinc/rails/errors/6960ad274ef4ae6af744a9c2
- bs#583e17 — `Rack::Multipart::EmptyContentError` — 1109×
- https://app.bugsnag.com/crimethinc/rails/errors/69372b4054690d304d583e17
- bs#36f108 — `ActionController::BadRequest` invalid encoding — 59×
- https://app.bugsnag.com/crimethinc/rails/errors/69a4d3b58d346d375136f108
- bs#e9652d — `Rack::QueryParser::InvalidParameterError` (`data:image/png;base64,...`) — 22×
- https://app.bugsnag.com/crimethinc/rails/errors/69e5b2be7aaa488f2ce9652d
- bs#c811b1 — `Rack::QueryParser::InvalidParameterError` SQLi probe (`SLEEP(5)`) — 14×
- https://app.bugsnag.com/crimethinc/rails/errors/69e10a178569dc55dec811b1
- `PG::NumericValueOutOfRange` — SQLi-probing bot sends `/page/-9491339 UNION ALL SELECT CONCAT(0x…) — -`. After existing `gsub(/\D/, "")` strips letters/symbols, the embedded digits in the SQL keywords + hex literals concatenate into a 36-digit number that overflows Postgres bigint when Kaminari calculates the OFFSET. (Was #5306; folded here.)
- https://app.bugsnag.com/crimethinc/rails/errors/69e408067aaa488f2c64cd60
[...]

PR #5323, opened 2026-04-30
[#5306] Clamp page param to avoid PG bigint overflow

### LINKED ISSUE #5306: [bs#64cd60] PG::NumericValueOutOfRange — params not validated before query
# bs#64cd60

https://app.bugsnag.com/crimethinc/rails/errors/69e408067aaa488f2c64cd60

`ActionView::Template::Error: PG::NumericValueOutOfRange: ERROR: value "56948034429339400356186387..." is out of range for type integer`. 14 occurrences.

A user-supplied param (likely an ID) is being passed straight to ActiveRecord without bounds checking. Validate / coerce the param at the controller before querying — out-of-range integers should 404, not 500.

veganstraightedge:
# #5306

https://app.bugsnag.com/crimethinc/rails/errors/69e408067aaa488f2c64cd60

## Summary

- `ArticlesController#index` was passing `params[:page]` to Kaminari without bounds. The existing `set_page_number` filter strips non-digits via `gsub(/\D/, '')` but does not clamp the magnitude.
- A SQLi-probing bot was sending paths like `/page/-9491339 UNION ALL SELECT CONCAT(0x7155656a6d726a,0x31,…)`. After `gsub`, the digits embedded in the SQL keywords and hex literals concatenate into a number with 36 digits — far larger than Postgres bigint can hold, so `Kaminari`'s `OFFSET = (page - 1) * per_page` overflowed and raised `PG::NumericValueOutOfRange` mid-render.
- Fix: extend `set_page_number` to clamp the cleaned page number to `PAGE_NUMBER_LIMIT = 1_000_000` (well within int4 bounds even after multiplying by `per_page`). When a clamp happens, `redirect_malformed_pagination` already redirects users to the corrected URL.
- Also switched the 2017 theme branch to use `@page_number` instead of `params[:page]` so json/atom feed requests (which skip `redirect_malformed_pagination`) get the same protection.

## Note

This will eventually be moot once #5303's bot/scraper Rack middleware lands and drops these probes before they reach Rails. Until then, controller-level clamping is cheap defense in depth.

## Test plan

- [ ] In staging, normal pagination still works on `/`, `/page/2`, etc.
- [ ] Visit `/page/569480344293394003561863877600713420` — should redirect to `/page/1000000` (or a similarly clamped value), not 500

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -5,6 +5,8 @@ class ArticlesController < ApplicationController
   before_action :redirect_malformed_pagination, only: :index
   before_action :force_2025_theme_for_feeds,    only: :index
 
+  PAGE_NUMBER_LIMIT = 1_000_000
+
   def index
     # TEMP: copied from home#index because the route /page/:page changed
     #       from home#index in 2017 theme
@@ -20,7 +22,7 @@ class ArticlesController < ApplicationController
       @latest_article = articles_for_current_page.first if params[:page].blank?
 
       # Feed artciles
-      @articles = articles_for_current_page.page(params[:page]).per(6)
+      @articles = articles_for_current_page.page(@page_number).per(6)
 
       render "#{Current.theme}/home/index"
     end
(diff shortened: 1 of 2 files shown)


veganstraightedge:
Closing — this is the wrong layer. The PG bigint overflow is a symptom of hostile bot traffic (SQLi probes hitting `/page/<huge-number>`), and the right fix is to drop those requests before they reach Rails. Folding into #5303 (Rack middleware to drop malformed bot/scraper requests). Branch left for reference; will not be merged.

PR #5325, opened 2026-04-30
[#5309] MusicController: redirect to Bandcamp, stop inheriting Tool flow

### LINKED ISSUE #5309: [bs#c4ca2c] tool_class.live — nil tool_class on tools#show
# bs#c4ca2c

https://app.bugsnag.com/crimethinc/rails/errors/69e89678ba7b406400c4ca2c

`NoMethodError: undefined method 'live' for nil` — `@tool = tool_class.live.published…`. 1 occurrence.

`tool_class` lookup is returning nil for an unknown tool slug/type. Should 404 instead of 500 when the requested tool type doesn't exist.

veganstraightedge:
# #5309

https://app.bugsnag.com/crimethinc/rails/errors/69e89678ba7b406400c4ca2c

## Summary

- `MusicController` was extending `ToolsController`, which runs `set_tool` / `set_tools` before_actions on every action. Both call `tool_class`, which returns `nil` when `controller_path` isn't in `ALL_TOOLS = %w[tools books zines journals posters stickers videos]`. `music` is not in that list (and there is no `Music` model to constantize), so `tool_class.live` raised `NoMethodError`.
- The original controller's comment was `# Redirects to Bandcamp` but the action was an empty `def index; end` — there was no view, no model, and no actual redirect. Music isn't a tool in the app's data model; it's a pointer to an external Bandcamp page.
- Fix: change `MusicController` to extend `ApplicationController` directly and actually redirect to https://crimethinc.bandcamp.com (which is what the site's footer / about copy already references for music).

## Test plan

- [ ] In staging, `/music` 302s to https://crimethinc.bandcamp.com

### DIFF
--- a/app/controllers/music_controller.rb
+++ b/app/controllers/music_controller.rb
@@ -1,4 +1,5 @@
-class MusicController < ToolsController
-  # Redirects to Bandcamp
-  def index; end
+class MusicController < ApplicationController
+  def index
+    redirect_to 'https://crimethinc.bandcamp.com', allow_other_host: true
+  end
 end
--- a/spec/controllers/music_controller_spec.rb
+++ b/spec/controllers/music_controller_spec.rb
@@ -1,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe MusicController do
+  describe 'GET #index' do
+    it 'redirects to the CrimethInc. Bandcamp page without raising' do
+      expect { get :index }.not_to raise_error
+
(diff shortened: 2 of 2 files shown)


just1602 (approved):
I'll assume it's AI generated, because this is not following the PR template and it's not your type of workflow to open draft PR and let them wait there... but gosh I don't understand why it put so much content in the PR summary, but create oo shitty commit message 😅 

But whatever, I don't really mind people using it as long as it doesn't generate a +10k / -10k lines review 😂

Issue #5349, opened 2026-05-06
Drop the CSS and JavaScript pre-processing dependencies


Bargraph6:
# What

We want to remove the following gems from the Gemfile:

-   `autoprefixer-rails`
-   `bootstrap`
-   `sassc-rails`
-   `uglifier`

And replace them with the `propshaft` Gem.

# Why

The original proof of concept (#5212) has more details, but the short
version is that having these dependencies:

1.  Makes setting up a development environment more difficult as a
    JavaScript engine is required.
2.  Leaves us more exposed to the JavaScript "supply chain", which is
    constantly being hacked.
3.  Increases the maintenance work as their are simply more dependencies

Now that CSS has many of the features of SCSS, there is not really a
reason to keep these pre-processors around.

Our JavaScript also doesn't need to be pre-processed, as there is very
little and it is mostly used as a progressive enhancement.

# How
## Demo

There was a proof-of-concept PR that layed out the rough gameplan.<sup><a id="fnr.1" class="footref" href="#fn.1" role="doc-backlink">1</a></sup>

| Demo of the site running with these dependencies dropped |
|--------|
| https://github.com/user-attachments/assets/a682c01c-e2da-41e8-bcf5-24a6b8f9e2c4 | 

## Implementation Plan

1.  Convert all of our `.scss` files to `.css`
    
    This involves replacing the SCSS syntax with CSS, then renaming the
    file. Ideally these are done in separate PRs so that the git
    history for the files is preserved on `main`.

2.  Vendor Bootstrap JS/CSS files
    
    As part of vendoring bootstrap we should be ablef to remove
    bootstrap gem.
    
    We can make use of this gem to help vendor bootstrap:
    
    -   <https://gist.github.com/veganstraightedge/ecca6887f60ada828a53ee71adcbf3a9>

3.  Replace `sprockets` with `propshaft`
    
    It is at this step where we will actually remove the gems:
    
    -   `autoprefixer-rails`
    -   `uglifier`
    -   `sassc-rails`
    
    and then add the gem:
    
    -   `propshaft`
    
    As part of setting up `propshaft`, we will nee to move the javascript
    code we have from `app/javascript` to `app/assets/javascripts` and
    remove `manifest.js`

# Changelog

Originally, I had been referencing the proof of concept PR in the PRs
related to this work. Going forward, I'll link back to this issue. To
make sure all of those past PRs have a link to this issue, they are
mentioned below. Future PRs will be linked in the normal github
[...]

PR #5375, opened 2026-05-14
css-refactor:(bootstrap) remove the bootstrap gem

Bargraph6:

* Gemfile: Remove the bootstrap gem.                                      
* Gemfile.lock: Ran `bundle install` to regenerate the lockfile.          
* lib/tasks/bootstrap.rake: copied the [rake task][0] from @veganstraightedge.
* lib/tasks/bootstrap.rake: removed the JS downloading and changed the `css_dir` to `assets/stylesheet/vendor`. Also added newline to version file to match existing file.

I ran the bootstrap.rake task, which resulted in no diff in `/assets`

[0]:https://gist.github.com/veganstraightedge/ecca6887f60ada828a53ee71adcbf3a9
                                                                        
- using the admin UI 
- delete the boostrap.css in assets/stylesheets/vendor
  - this will break admin
- run the rake task
  - this will fix admin
  - there should be no diff when running `git diff main..HEAD -- app/assets`
    
As part of the effort to drop the CSS/JS preprocessor dependencies, we    
will start vendoring bootstrap directly rather than going through a       
gem.                                                                      
                                                                          
Here are my operating assumptions, (which may be wrong!):                 
                                                                          
1. We only use bootstrap CSS, not JS                                      
2. We still want the vendored CSS in `app/assets/stylesheets/vendor`      
   - as opposed to `vendor/stylesheets`                                   
                                                                          
The overview of this change is that I:                                    
                                                                          
1. removed the bootstrap gem                                              
2. copied the [rake task][0] from @veganstraightedge into the project.   
  - removed the JS bits from the task                                     
  - changed the download dir to match the existing dir                    
3. deleted `app/assets/stylesheets/vendor/bootstrap*`                     
  - at this point, admin is broken                                        
4. ran the rake task                                                      
  - at this point, admin works                                            
  - no diff in the assets as we downloaded the same bootstrap version     

# Acceptance Criteria

## Questions for the reviewer

---                                                                       
                                                                          
- related-to: https://github.com/crimethinc/website/issues/5349


lib/tasks/bootstrap.rake
```diff
@@ -0,0 +1,31 @@
...
+    response       = HTTP.get 'https://api.github.com/repos/twbs/bootstrap/releases/latest'
+    latest         = JSON.parse response.body.to_s
+    latest_version = latest['tag_name'].delete_prefix 'v'
+
+    puts "updating from #{current_version} to #{latest_version}"
+
+    css_url     = "https://cdn.jsdelivr.net/npm/bootstrap@#{latest_version}/dist/css/bootstrap.css"
+    css_map_url = "https://cdn.jsdelivr.net/npm/bootstrap@#{latest_version}/dist/css/bootstrap.css.map"
+
+    css_dir = Rails.root.join 'app/assets/stylesheets/vendor'
```
Bargraph6, line 21: @veganstraightedge @just1602 this is ready for review.

Maybe we want to move the download location up to `/vendor/stylesheets', but for now I just went with the existing location so that I am just changing one thing at a time (how the file is downloaded vs how it is downloaded *and* how it is served in the app).

just1602: We might want to change the vendoring folder, but for now it seems good to me!

just1602 (approved):

just1602:
We should look if we use JS in the admin for a humberger menu or anything like that, but otherwise it seems really good!

Bargraph6:
> We should look if we use JS in the admin for a humberger menu or anything like that, but otherwise it seems really good!

@just1602 oh, you are right. In the admin:

https://github.com/crimethinc/website/blob/41ca76c32cc70d6f43163f04d947ee890f27144c/app/views/layouts/admin.html.erb#L36-L40

I guess I never noticed since I block JS by default. On the plus side, this is not loading from the vendor directory, so this change shouldn't change anything.

I could either:

- update this PR to put back in the JS downloading in @veganstraightedge's rake task, then change the admin to use the vendor version
- rework the admin to not use bootstrap for a hamburger menu (either write the clickhandler myself or even just do it in pure css with a popover  + anchor positioning :thinking: )

just1602:
If you can do it without js, it's probably the best.

Otherwise, if it doesn't affect admin nor the publish now button we can merge.

PR #5398, opened 2026-05-23
spec: refactor user_spec

just1602:

Refactor user_spec to use let and subject

`bin/rspec` and `bin/rubocop` shouldn't fail.

We don't have a total coverage, and some specs are older / are probably not following the same pattern as the most recent one, so this try to unify the specs.

# Acceptance Criteria

## Questions for the reviewer

---

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,52 +1,60 @@
 require 'rails_helper'
 
 RSpec.describe User do
-  describe '#name' do
-    it 'returns the username' do
-      user = described_class.new(username: 'example')
+  subject(:user) { user }
 
-      expect(user.name).to eq 'example'
+  describe '#name' do
+    describe 'returns the username' do
+      subject { user.name }
+
+      let(:user) { build(:user, username: 'example') }
+
+      it { is_expected.to eq 'example' }
     end
   end
 
   describe '#strip_whitespace' do
-    it 'strips leading and trailing whitespace from username' do
-      user = create(:user, username: '  padded  ', password: 'x' * 30)
+    context 'when there is leading and trailing whitespace in a username' do
+      subject { user.username }
 
-      expect(user.username).to eq 'padded'
+      let(:user) { create(:user, username: '  padded  ') }
+
+      it { is_expected.to eq 'padded' }
     end
   end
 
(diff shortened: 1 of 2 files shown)


just1602:
@Bargraph6 if you also have a sec and want to give a look at this one, I'm open to all your comments.

spec/models/user_spec.rb
```diff
@@ -1,52 +1,62 @@
...
-    it 'rejects passwords shorter than minimum length' do
-      user = build(:user, password: 'short')
-
-      expect(user).not_to be_valid
-      expect(user.errors[:password]).to be_present
+    context 'when passwords shorter than minimum length' do
+      let(:user) { build(:user, password: 'short') }
+
+      it 'is expected to be invalid and properly set the error message' do
+        # NOTE: we need to call `#valid?` on the model for the error message
```
Bargraph6: We could probably just remove this comment.

The spec will fail if someone tries splitting in the future without pre-calling .valid? somewhere

Bargraph6 (approved):
@just1602 this lgtm!

spec/models/user_spec.rb
```diff
@@ -1,52 +1,60 @@
...
 
-    it 'does not allow authors to publish' do
-      user = build(:user, role: 'author')
+    context 'when the user is an author' do
+      let(:user) { build(:user, :author) }
 
-      expect(user.can_publish?).to be false
-      expect(user.cant_publish?).to be true
+      it { expect(user.can_publish?).to be false }
+      it { expect(user.cant_publish?).to be true }
```
veganstraightedge, line 40: Overall this PR is an improvement to the codebase! Thank you.

---

Aside, for future PR. Just thinking out loud here.

These couple tests remind that we should try to find names for predicate methods that work better with rspec predicate matchers.

```ruby
expect(user.can_publish?).to be false 
expect(user).not_to be_can_publish
# equivalent behavior but obviously the second reads badly 
```

```ruby
# made up example, for "`user.is_purple?`"
# bad
expect(user.is_purple?).to be true
# good
expect(user).to be_purple
# ^^^ that's prettier! ^^^
```

veganstraightedge (approved):

PR #5400, opened 2026-05-24
ci: fix failling build with github ubuntu 24.04

just1602:

CI was failing for a couple of day due to an archive not being found in the archive. Simply running `apt-get update` before installation the packages fix the issue.

It takes a little bit more time, but I also added `--no-install-recommands` which reduce the amount of packages installed and should compensate for the time spent updating the package repository.

CI should pass.

??

# Acceptance Criteria

## Questions for the reviewer

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - name: Install system libraries
-        run: sudo apt-get -y install libpq-dev libvips-dev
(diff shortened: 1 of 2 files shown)


just1602:
@veganstraightedge I'll need your review on this one.

just1602:
GitHub as updated the base image, so CI pass again, but I still think we should at least keep the part where we do not install recommended packages.

veganstraightedge (approved):

